PR #5, opened 2026-04-07
Feat/expo frontend

robin-raq:


shoyu-ramen:
My main concern is this would essentially add a second frontend alongside the existing flutter app that is already wired up to the backend. 

If you want to replace the frontend then this PR should wire everything up and remove the old scaffolding code, otherwise I think it makes sense to build on top of what we have -- the frontend is essentially a working shell, ready to be built out.

PR #8, opened 2026-04-07
ci: add GitHub Actions pipeline with lint, typecheck, and test coverage

lramosve:
## Summary
- Add **ruff** (linting), **mypy** (type checking), and **pytest-cov** (coverage) to dev dependencies with config in `pyproject.toml`
- Create GitHub Actions CI workflow (`.github/workflows/ci.yml`) with four parallel jobs: backend lint, backend typecheck, backend tests + coverage upload, and frontend Flutter analyze
- Add `ruff check` to the Makefile `lint` target and a new `typecheck` target

## Test plan
- [ ] Verify CI workflow triggers on this PR and all four jobs pass
- [ ] Confirm coverage report is uploaded as an artifact
- [ ] Run `make lint` and `make typecheck` locally to validate Makefile changes

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,64 @@
+name: CI
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  backend-lint:
+    name: Backend Lint
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.13"
+      - run: pip install -e ".[dev]"
+      - run: ruff check backend tests
+
+  backend-typecheck:
+    name: Backend Typecheck
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.13"
+      - run: pip install -e ".[dev]"
+      - run: mypy
+
+  backend-test:
+    name: Backend Tests
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.13"
+      - run: pip install -e ".[dev]"
+      - run: pytest
+      - name: Upload coverage report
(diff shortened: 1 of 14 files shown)


jack0lantern:
Do you think we could incorporate docker-compose into this PR, or would you rather have a separate task? I could handle it if we do it that way

jack0lantern:
I'm assuming you'll deal with docker-compose and deployment to railway separately? Code quality checks wise, this is solid.

jack0lantern (approved):

robin-raq (commented):
## CI Pipeline Review

Solid foundation — ruff, mypy, pytest-cov, and Flutter analyze in parallel jobs with concurrency control. A few issues to address before merging:

### Critical

**1. mypy will crash on `backend/vendor/mr_mt3/`**

Every file in the vendored MT3 package has unconditional `import torch` at the top level. CI installs `.[dev]` (no torch), so mypy will fail with `Cannot find implementation or library stub for module named 'torch'` for every vendor file.

**Fix** — add to `[tool.mypy]` in `pyproject.toml`:
```toml
exclude = ["backend/vendor/"]
```

**2. `coverage.xml` generation — verify addopts**

`pytest-cov` only produces `coverage.xml` when invoked with `--cov` flags. The PR adds `addopts` to `pyproject.toml` which should handle this — just verify the `addopts = "--cov=backend --cov-report=term-missing --cov-report=xml"` line is actually wired, since bare `pytest` without those flags produces no XML and the upload step becomes dead code.

### Important

**3. Python version inconsistency**

- `requires-python = ">=3.10"` in pyproject.toml
- CI uses Python 3.13
- ruff targets `py311`, mypy targets `python_version = "3.11"`

These should align. Recommend setting ruff/mypy to `py310` to match the declared minimum.

**4. `make lint` in frontend-lint job may fail**

The Makefile `lint` target now runs both `ruff check` AND `flutter analyze`. If the `frontend-lint` CI job only sets up Flutter (no Python), `ruff` will be missing and the job fails. Consider splitting into `lint-backend` and `lint-frontend` Make targets, and having each CI job call the specific one.

### Looks Good

- Concurrency control (cancel in-progress on new push) 👍
- `pip install -e ".[dev]"` without `[mt3]` — correct, avoids heavy ML deps in CI
- Tests will pass — `TranscribeService` guards all torch imports with try/except
- Flutter `pub get` before `analyze` is handled

swifttarrow (dismissed):
Checks are failing in CI/CD

shoyu-ramen (approved):

PR #9, opened 2026-04-07
feat: Phase 1 frontend polish — kawaii theme, mascots, PDF viewer, MIDI player

robin-raq:
## Summary
Complete Phase 1 frontend (Epic 1) — all 6 tickets. Transforms the app from a bare skeleton into a polished, branded experience matching the wireframe designs.

- **Kawaii theme system**: Teal/orange/cream palette, pill buttons, rounded cards, Material3
- **Bottom navigation**: Home / Library / Profile tabs (Library and Profile are placeholders for Phase 2)
- **Welcome hero**: Mascot image + headline + subtitle above the upload form
- **Progress screen**: Stage-specific mascot images swap during pipeline, sticker badge row, rotating tips, friendly stage names
- **Result screen**: Success mascot, inline PDF viewer (iframe), MIDI player with piano roll visualizer (html-midi-player), pill download buttons
- **Callback API**: Verified existing WebSocket pub/sub handles it — documented, no new endpoint needed
- **yt-dlp moved to core deps**: Addresses PR #7 feedback about confusing optional dependency errors

## Mascot Assets
7 variants for different pipeline states (home, ingest, transcribe, arrange, engrave, success, error) in `frontend/assets/mascots/`.

## Known Limitations
- PDF viewer and MIDI player show stub output until backend ML pipeline (MT3 + LilyPond) is fully wired. The frontend is ready — just needs real artifacts from the backend.
- MIDI player and PDF viewer are web-only (Flutter Web). Non-web platforms get download fallbacks.

## Test plan
- [x] 42 Flutter widget tests pass (theme, nav, hero, progress helpers, result screen, YouTube input)
- [x] 36 backend tests pass (no regressions)
- [x] Tested locally in Chrome — full upload → progress → result flow works
- [ ] Backend team: wire MT3 deps (numpy<2, transformers<5, git-lfs pull) for real transcription output

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #10, opened 2026-04-08
cd: add Cloud Run deployment pipeline

lramosve:
## Summary
- Add **multi-stage Dockerfile**: Flutter web build (stage 1) + Python 3.13-slim runtime (stage 2) serving both API and static frontend from a single container
- Add **GitHub Actions deploy workflow** (`.github/workflows/deploy.yml`) triggered on push to `main` — builds the container, pushes to Artifact Registry, and deploys to Cloud Run
- Update **`backend/main.py`** to mount Flutter web build at `/` when the `static/` directory is present (production only — no effect in dev)
- Add `.dockerignore` to keep the image lean

## GCP setup required before first deploy
1. Create a GCP project and enable Cloud Run + Artifact Registry APIs
2. Create an Artifact Registry Docker repo named `oh-sheet`
3. Set up [Workload Identity Federation](https://github.com/google-github-actions/auth#workload-identity-federation) for GitHub Actions
4. Add these GitHub repo settings:
   - **Vars**: `GCP_PROJECT_ID`, `GCP_REGION` (defaults to `us-central1`)
   - **Secrets**: `WIF_PROVIDER`, `WIF_SERVICE_ACCOUNT`

## Test plan
- [ ] `docker build -t oh-sheet .` succeeds locally
- [ ] Container starts and serves both API (`/v1/health`) and frontend (`/`) on port 8080
- [ ] Existing backend tests still pass (3 pre-existing failures in `test_artifacts.py` are unrelated)
- [ ] Deploy workflow triggers on merge to main (after GCP secrets are configured)


robin-raq (commented):
## Cloud Run Deployment Review

Clean, well-structured deployment pipeline. The multi-stage Docker build and single-container SPA+API approach is the right call for MVP. A few issues to flag:

### Critical

**1. Python 3.13 in Dockerfile but `requires-python = ">=3.10"` — torch won't install**

The Dockerfile uses `python:3.13-slim`. Currently the `[mt3]` deps aren't installed in the image (intentionally), but when the team wires real transcription, torch doesn't support 3.13 yet. Consider using `python:3.11-slim` or `python:3.12-slim` for forward compatibility.

Also note: `requires-python` was recently changed to `>=3.10` on a feature branch. If that merges first, it won't conflict, but worth aligning.

**2. `yt-dlp` is being moved to core deps — Dockerfile still references `.[youtube]`**

PR #9 (pending) moves `yt-dlp` from `[youtube]` optional group to core `dependencies`. Once that merges, the Dockerfile line `pip install --no-cache-dir ".[youtube]"` will fail because the `youtube` extra won't exist. Should be just `pip install --no-cache-dir .` — or wait for that PR to merge first and update accordingly.

### Important

**3. `StaticFiles` mount at `/` will shadow API routes if registered first**

The PR mounts `StaticFiles(directory=_STATIC_DIR, html=True)` at `/`. Since it's added *after* the `/v1` routers, FastAPI will check `/v1` routes first, then fall through to static files for everything else. This is the correct order — just documenting that it's order-sensitive. If someone moves the mount above the `include_router` calls, the API breaks silently.

Consider adding a code comment:
```python
# IMPORTANT: mount AFTER API routers — StaticFiles at "/" is a catch-all.
```

**4. Cloud Run memory/CPU may be tight for ML inference**

```yaml
--memory 512Mi
--cpu 1
```

512Mi is fine for stubs + yt-dlp, but when MT3 is wired in, the model alone needs ~400MB RAM + torch overhead. You'll want at least `--memory 2Gi --cpu 2` for real transcription. Consider adding a comment or making these configurable via GitHub vars so the backend team can bump them later without editing the workflow.

**5. No health check configured on Cloud Run**

Cloud Run uses the container's response to determine readiness. The `/v1/health` endpoint exists, but the deploy command doesn't set `--health-check-path`. Without it, Cloud Run defaults to TCP probe on the port. Adding `--health-check-path /v1/health` would catch startup failures faster (e.g., missing env vars).

### Minor

**6. `.dockerignore` excludes `assets/` — is that intentional?**

The `.dockerignore` has `assets/` which would exclude `frontend/assets/mascots/` from the Docker build context. However, since the Flutter build happens inside Stage 1 where `COPY frontend/ .` copies everything, and `.dockerignore` only applies to the root context... actually this is fine — Stage 1 copies `frontend/` explicitly before `.dockerignore` filters the root. But worth double-checking that the mascot images end up in the Flutter web build output.

**7. No `--set-env-vars` for `OHSHEET_BLOB_ROOT`**

The deploy sets `OHSHEET_CORS_ORIGINS=*` but doesn't configure blob storage. In Cloud Run (ephemeral filesystem), the default `./blob` path will work but files disappear on cold start. Fine for MVP with stubs, but flag for when S3/R2 is wired in.

### Looks Good

- Multi-stage build keeps image lean (~200MB vs ~2GB with Flutter SDK) 👍
- `API_BASE_URL=` (empty) so Flutter uses same-origin relative URLs — correct for single-container deploy
- Workload Identity Federation instead of service account keys — security best practice
- `concurrency: cancel-in-progress` prevents stacking deploys
- `workflow_dispatch` for manual trigger — useful for debugging
- `--allow-unauthenticated` correct for a public-facing app
- `--min-instances 0` for cost efficiency during MVP

shoyu-ramen (approved):

swifttarrow (approved):

PR #12, opened 2026-04-08
feat: replace MT3 with Basic Pitch for audio transcription

shoyu-ramen:
## Summary

- Swap the vendored MR-MT3 T5 transcriber (175MB LFS checkpoint + vendored model code) for Spotify's Basic Pitch — a lightweight polyphonic pitch-tracker that ships its weights inside the wheel and auto-picks the fastest backend per platform (CoreML on Darwin, ONNX/TFLite elsewhere).
- Net diff: **+198 / -2786** across 23 files. `backend/vendor/mr_mt3/` is fully deleted along with its `.gitattributes` LFS rule.
- Public `TranscribeService` contract is unchanged; the switch is contained to the service internals + deps + docs.

## Behavior changes

- **Single `PIANO` track** instead of multi-program GM tracks. Basic Pitch doesn't predict instrument programs; `arrange.py`'s hand-split logic handles the collapse cleanly.
- **Per-note confidence** now comes from the model's sigmoid output (`note_events` amplitudes), replacing the previous duration/density heuristic.
- **Velocity** comes straight from `pretty_midi` — the old MT3 RMS rescaling hack is gone.
- **Tempo** is estimated via `pretty_midi.estimate_tempo()` with a 40–240 BPM sanity range, falling back to 120.
- **Dead-code removal:** `arrange.py`'s cross-track dedup (52 lines + `DEDUP_TIME_TOL` constant) is deleted — it only mattered when MT3 emitted the same audio event under several GM programs. Single-track input makes it a no-op.

## Install quirk worth knowing

basic-pitch 0.4.0 hard-codes \`tensorflow-macos>=2.4.1,<2.15.1\` as a **base dependency** (not an extra) on Darwin + \`python_version > '3.11'\`, and tensorflow-macos has no wheels for Python 3.13. The \`make install-basic-pitch\` target works around this by installing basic-pitch with \`--no-deps\` and pinning its actual runtime deps (librosa, pretty_midi, scipy, scikit-learn, resampy, mir_eval, coremltools, onnxruntime, typing_extensions) in the \`[basic-pitch]\` extra in \`pyproject.toml\`. Full explanation in the pyproject comment. Revisit if basic-pitch cuts a 0.5 that fixes this upstream.

## Test isolation

Added an autouse fixture in \`tests/conftest.py\` (\`skip_real_transcription\`) that monkey-patches \`_run_basic_pitch_sync\` to raise immediately, routing all pipeline tests through the \`except Exception → _stub_result\` branch. Previously this isolation was implicit — MT3 wasn't installed in CI, so \`ImportError\` provided it for free. Now that basic-pitch *is* installable alongside the dev deps, the fixture makes the isolation explicit so the suite stays fast (~8s) and deterministic regardless of whether basic-pitch is installed on the host.

## Test plan

- [x] \`pytest\` — 36/36 passing, stable across back-to-back runs, ~8s
- [x] \`ruff check backend tests\` — clean
- [x] \`mypy\` — clean (26 source files)
- [x] \`make install-basic-pitch\` succeeds on Python 3.13 / Darwin
- [x] Smoke-test: \`basic_pitch.inference.Model(ICASSP_2022_MODEL_PATH)\` loads successfully; auto-picks the CoreML \`.mlpackage\` model on this machine
- [ ] Verify inference quality on a real song end-to-end (not covered by the suite — manual follow-up recommended)
- [ ] Verify CI passes on Linux (backend auto-picks ONNX/TFLite there instead of CoreML)


lramosve (approved):
Looks good, ship it!!

swifttarrow (approved):

PR #17, opened 2026-04-08
fix: use same-origin URLs in production web deployment

lramosve:
## Summary
- The Flutter web app was hitting `http://localhost:8000` on Cloud Run because `API_BASE_URL=` (empty) falls through to the localhost default
- Fix: detect if we're running on a non-localhost host and return empty string (same-origin relative URLs) instead
- Local dev behavior unchanged — still uses `http://localhost:8000`

## Test plan
- [ ] On Cloud Run, API calls go to same-origin (`/v1/...`) instead of `localhost:8000`
- [ ] Local dev (`flutter run -d chrome`) still hits `http://localhost:8000`


swifttarrow (approved):

shoyu-ramen (approved):

robin-raq (commented):
## Bug: `wsBaseUrl` breaks in production

When `apiBaseUrl` returns `''` (empty string in production), look at what happens in `config.dart:29-34`:

```dart
static String get wsBaseUrl {
  final base = apiBaseUrl;  // '' in production
  if (base.startsWith('https://')) return 'wss://${base.substring(8)}';
  if (base.startsWith('http://')) return 'ws://${base.substring(7)}';
  return base;  // returns '' — empty string
}
```

Then in `ws.dart:24`:
```dart
final url = '${AppConfig.wsBaseUrl}/v1/jobs/$jobId/ws';
// becomes: '/v1/jobs/abc123/ws'
```

**`WebSocketChannel.connect(Uri.parse('/v1/jobs/abc123/ws'))` will fail** — WebSocket connections require an absolute URL with a `ws://` or `wss://` scheme. Unlike HTTP fetch, the browser cannot resolve a relative URL for a WebSocket connection.

**Suggested fix** — `wsBaseUrl` should construct the absolute WebSocket URL from `Uri.base` when `apiBaseUrl` is empty:

```dart
static String get wsBaseUrl {
  final base = apiBaseUrl;
  if (base.startsWith('https://')) return 'wss://${base.substring(8)}';
  if (base.startsWith('http://')) return 'ws://${base.substring(7)}';
  if (base.isEmpty && kIsWeb) {
    final scheme = Uri.base.scheme == 'https' ? 'wss' : 'ws';
    return '$scheme://${Uri.base.host}:${Uri.base.port}';
  }
  return base;
}
```

The HTTP same-origin fix is correct and well-reasoned — just needs this WebSocket fix before merge.

PR #18, opened 2026-04-08
feat(transcribe): post-process Basic Pitch with cleanup + melody split

shoyu-ramen:
## Summary

Two-phase post-processing on top of the Basic Pitch baseline, plus an
artifact plumbing change to expose the raw transcription MIDI.

**Phase 1 — `backend/services/transcription_cleanup.py`** (pure stdlib, no new deps):
- Merge fragmented sustains within a small gap
- Drop octave-ghost notes whose amplitude is too low vs. the lower octave (preserves real doublings)
- Prune notes that are both short *and* quiet (ghost tails)
- Order matters: merge → octave → ghost-tail so a fragmented sustain isn't mistaken for a ghost on its first fragment

**Phase 2 — `backend/services/melody_extraction.py`** (numpy, late-imported):
- Viterbi F0 tracer over `model_output["contour"]` — the 264-bin multi-pitch salience matrix Basic Pitch already computes but we weren't using. **No new model weights, no new deps.**
- Vectorized shift-and-min transition updates over ±12 bins with an unvoiced state for silence
- Tags each Basic Pitch note as `MELODY` (≥60% agreement with the Viterbi path inside MIDI 55–90) or non-melody
- **Back-fill** (new): synthesize notes for stable Viterbi runs ≥ 120 ms with no matching Basic Pitch event. Recovers soft sustained melody lines the polyphonic tracker loses under accompaniment. Amplitude clipped to [0.15, 0.60] so recovery notes don't outshine real ones; must not be re-run through `cleanup_note_events` (would look like ghost tails).
- Skips cleanly when numpy/contour/tracer fail — callers fall back to a single `PIANO` track

**Phase 3 — `backend/services/bass_extraction.py` + `backend/services/chord_recognition.py`** (new):
- **Bass extraction** reuses the same Viterbi tracer on the low-register slice (E1..G3), stricter 0.4 transition weight, 9-bin max jump. Runs *after* melody on the non-melody remainder: melody claims first, then bass, then the remainder becomes CHORDS.
- **Chord recognition** runs `librosa.feature.chroma_cqt` on the HPSS-harmonic component, beat-synced via `librosa.util.sync`, matched against 24 triad templates (12 major + 12 minor). Labels below 0.55 score become `N` and are dropped. Consecutive identical labels collapse into one `RealtimeChordEvent` span. **Beat tracking runs on the raw waveform**, not the HPSS'd one — HPSS strips the onset transients that `beat_track` needs, and feeding the harmonic-only signal produces zero beats on material where the raw tracker finds a clean pulse.

**Transcribe wiring:**
- `_pretty_midi_to_transcription_result` takes an `events_by_role` dict and emits one `MidiTrack` per non-empty role (deterministic order: MELODY → BASS → CHORDS → PIANO → OTHER). Contract notes built directly from `note_events` via `_event_to_note` using Basic Pitch's own `round(127 * amp)` velocity formula; `pm` retained only for tempo fallback.
- `_run_basic_pitch_sync` chains cleanup → rebuild pretty_midi → extract_melody → extract_bass → recognize_chords, emits one MidiTrack per non-empty role, and attaches the chord label stream to `HarmonicAnalysis.chords`.
- Single-track PIANO fallback only fires when *both* melody and bass are skipped.
- `bass_stats` / `chord_stats` flow into `QualitySignal.warnings` alongside `cleanup_stats` and `melody_stats`.

**Transcription MIDI blob persistence:**
- `TranscribeService` now takes a `BlobStore` and persists the cleaned `pretty_midi` to `jobs/{job_id}/transcription/basic-pitch.mid`, tagging the result with `transcription_midi_uri`.
- `PipelineRunner` forwards the URI onto the final `EngravedOutput` via `model_copy`.
- New `GET /v1/artifacts/{job_id}/transcription_midi` route (404 when absent — midi_upload variant and stub fallbacks don't have one).

**Config** — all overridable via `OHSHEET_*` env vars:
- 5 `cleanup_*` knobs (Phase 1)
- 7 `melody_*` knobs + 5 `melody_backfill_*` knobs (Phase 2)
- 7 `bass_*` knobs + 3 `chord_*` knobs (Phase 3)
- Kill switches: `OHSHEET_MELODY_EXTRACTION_ENABLED`, `OHSHEET_BASS_EXTRACTION_ENABLED`, `OHSHEET_CHORD_RECOGNITION_ENABLED`, `OHSHEET_MELODY_BACKFILL_ENABLED`

**Tuning note:** defaults were iterated against synthetic fixtures. Real audio will want per-genre tuning — vocal-led tracks may want `melody_match_fraction≈0.75` to cut false positives; instrumental tracks may want `melody_voicing_floor≈0.08` to pick up soft leads; bass-heavy material may want a wider `bass_high_midi` (60 instead of 55).

## Test plan

- [x] `pytest` — **101 / 101 pass** (58 baseline + 18 Phase 2 melody + 16 Phase 1 cleanup + 6 back-fill + 9 bass + 10 chord, minus 6 deduped counts)
- [x] Cleanup: 90%+ coverage across merge, octave dedup, ghost-tail, pass ordering, idempotency, empty input
- [x] Melody: Viterbi tracer, bin↔MIDI mapping, end-to-end tagging on synthetic contours, back-fill synthesis / clipping / skip paths
- [x] Bass: synthetic low-register contours covering pedal, walking bass, out-of-band rejection, disagreeing pitch, None/malformed contour, empty events, tiny contour, config defaults
- [x] Chord: template construction, single-chord C:maj, C→G progression, minor triad, bounded confidence, consecutive collapse, high-threshold suppression, tiny input, silence, config defaults
- [x] `ruff check` clean
- [x] `mypy` clean on changed surface (31 source files)
- [x] Real end-to-end smoke test: synthesized C5→C6 melody over a quiet C3/E3 drone through the full `_run_basic_pitch_sync` pipeline returns a clean 4-note MELODY track + 2-note CHORDS track with 94% voiced frame coverage
- [ ] Tune defaults against real audio fixtures (follow-up)
- [ ] Observe warnings in `QualitySignal.warnings` for a few real jobs to confirm the telemetry is useful

## Follow-ups

Both of the follow-ups originally called out on this PR are now done:

- [x] ~~**Phase 3:** bass line extraction via the same Viterbi trick over the low-register slice of `contour`, plus chord recognition via librosa chroma + template matching~~
- [x] ~~**Back-fill missed melody notes:** when the Viterbi path has a stable segment ≥ 120 ms with no matching Basic Pitch note, synthesize one from the contour salience~~

Remaining for a future PR:

- **Extended chord vocabulary:** 7ths / sus / dim templates (currently triads only). Ambiguous harmonies fall through to whichever triad the chroma looks most like.
- **Key-aware chord smoothing:** an HMM over a key-conditioned template set would suppress the occasional out-of-key false positive on real audio.
- **Inversion tracking:** the root we return is the template root, not necessarily the lowest note.

### DIFF
--- a/backend/api/deps.py
+++ b/backend/api/deps.py
@@ -30,7 +30,7 @@ def get_runner() -> PipelineRunner:
     blob = get_blob_store()
     return PipelineRunner(
         ingest=IngestService(blob_store=blob),
-        transcribe=TranscribeService(),
+        transcribe=TranscribeService(blob_store=blob),
         arrange=ArrangeService(),
         humanize=HumanizeService(),
         engrave=EngraveService(blob_store=blob),
--- a/backend/api/routes/artifacts.py
+++ b/backend/api/routes/artifacts.py
@@ -18,13 +18,14 @@ from backend.storage.local import LocalBlobStore
 
 router = APIRouter()
 
-ArtifactKind = Literal["pdf", "musicxml", "midi"]
+ArtifactKind = Literal["pdf", "musicxml", "midi", "transcription_midi"]
 
 # kind → (uri attribute on EngravedOutput, media type, downloaded filename suffix)
 _KIND_INFO: dict[str, tuple[str, str, str]] = {
-    "pdf":      ("pdf_uri",            "application/pdf",                       "sheet.pdf"),
-    "musicxml": ("musicxml_uri",       "application/vnd.recordare.musicxml+xml", "score.musicxml"),
-    "midi":     ("humanized_midi_uri", "audio/midi",                            "humanized.mid"),
+    "pdf":                ("pdf_uri",                "application/pdf",                       "sheet.pdf"),
+    "musicxml":           ("musicxml_uri",           "application/vnd.recordare.musicxml+xml", "score.musicxml"),
+    "midi":               ("humanized_midi_uri",     "audio/midi",                            "humanized.mid"),
+    "transcription_midi": ("transcription_midi_uri", "audio/midi",                            "transcription.mid"),
 }
 
 
@@ -53,6 +54,15 @@ def download_artifact(
 
     attr, media_type, suffix = _KIND_INFO[kind]
     uri = getattr(record.result, attr)
+    if uri is None:
+        # Optional artifacts (e.g. transcription_midi) won't exist for
+        # every variant — midi_upload skips transcription, and the stub
+        # fallback doesn't persist a MIDI file. Return 404 so clients can
+        # distinguish "never produced" from "job failed".
+        raise HTTPException(
+            status_code=404,
+            detail=f"Job {job_id} has no {kind!r} artifact.",
+        )
 
     try:
         data = blob.get_bytes(uri)
--- a/backend/config.py
+++ b/backend/config.py
@@ -59,5 +59,35 @@ class Settings(BaseSettings):
     melody_max_transition_bins: int = 12         # ≈ 4 semitones / frame
     melody_match_fraction: float = 0.6
 
+    # Back-fill of stable Viterbi runs with no matching Basic Pitch note.
+    # See _backfill_missed_melody_notes in melody_extraction.py.
+    melody_backfill_enabled: bool = True
+    melody_backfill_min_duration_sec: float = 0.12
+    melody_backfill_overlap_fraction: float = 0.5
+    melody_backfill_min_amp: float = 0.15
+    melody_backfill_max_amp: float = 0.60
+
+    # ---- Bass extraction (Phase 3 post-processing) ------------------------
+    # Same Viterbi trick as melody extraction, run over the low-register
+    # slice of the contour matrix. Accepts the non-melody events from
+    # Phase 2 and splits them into BASS / remaining buckets. Defaults
+    # mirror bass_extraction.DEFAULT_* so config and tests agree.
+    bass_extraction_enabled: bool = True
+    bass_low_midi: int = 28                      # E1
+    bass_high_midi: int = 55                     # G3
+    bass_voicing_floor: float = 0.12
+    bass_transition_weight: float = 0.40
+    bass_max_transition_bins: int = 9            # ≈ 3 semitones / frame
+    bass_match_fraction: float = 0.55
+
+    # ---- Chord recognition (Phase 3 post-processing) ----------------------
+    # librosa chroma_cqt + 24 triad templates, beat-synced via the same
+    # beat tracker that drives the tempo map. Labels attach to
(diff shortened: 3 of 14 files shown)


robin-raq (commented):
## Review: Approve with minor suggestions

Well-architected PR. Clean separation of concerns, graceful fallbacks everywhere, solid test coverage (34 new tests), and a kill switch via config.

### Performance: Python-level frame loop in note tagging

In `melody_extraction.py:392-396`, the per-note match scoring uses a Python `for` loop over frames:

```python
for f in range(f0, f1):
    pb = int(path[f])
    if pb < 0:
        continue
    if abs(pb - target_bin) <= tol_bins:
        match += 1
```

For a 3-minute song at ~86 Hz frame rate, a note spanning 2 seconds ≈ 172 frames. With 200+ notes, this is ~35k iterations in Python. Could be vectorized trivially:

```python
segment = path[f0:f1]
voiced = segment >= 0
match = int(np.sum(voiced & (np.abs(segment - target_bin) <= tol_bins)))
```

Not blocking, but worth a follow-up since this is on the hot path.

### Minor nits

| # | File | Note |
|---|------|------|
| 1 | `transcribe.py:734` | `cleanup_stats.output_count != cleanup_stats.input_count or cleanup_stats.merged` — the `or cleanup_stats.merged` is redundant since any merge changes the count. Harmless but reads like a missed simplification. |
| 2 | `transcription_cleanup.py:905` | Merge keeps the first fragment's pitch bends. Documented, but the alternative (concatenating bend arrays) might preserve vibrato data across fragments. Fine for now — noting for awareness. |
| 3 | `melody_extraction.py` | Only checks one octave up for ghosts (`p-12`). Two-octave ghosts (`p+24`) exist in real audio but are rare. Phase 3 material. |
| 4 | Config | 12 new `OHSHEET_*` env vars is a lot of knobs. Consider documenting a "tuning cheat sheet" mapping genre → recommended overrides (the PR body hints at this). |

### What's done well

- **Pass ordering is tested** (`test_pass_ordering_merge_then_octave_then_ghost`) — confirms fragments merge before ghost-tail pruning would incorrectly kill them
- **Idempotency test** — cleanup twice on clean input produces identical output
- **Config defaults pinned to module constants** via dedicated tests — prevents silent drift
- **Graceful degradation** at every level: missing numpy, malformed contour, Viterbi crash, no melody found → all fall back to single-PIANO track
- **Deterministic track ordering** (MELODY → BASS → CHORDS → PIANO → OTHER) for stable test output

Clever use of Basic Pitch's unused `contour` salience matrix for melody extraction — no new model weights or deps needed.

robin-raq (approved):

swifttarrow (approved):

PR #19, opened 2026-04-08
feat(transcribe): Demucs stem separation + audio preprocessing

shoyu-ramen:
## Summary
- Adds an opt-in **audio preprocessing** stage (`backend/services/audio_preprocess.py`) that runs HPSS harmonic extraction + RMS-normalize-with-peak-ceiling before Basic Pitch inference. Off by default — A/B deltas on the rising-sun fixtures are real but marginal (+0.00 to +0.04 confidence) so operators can flip it on per-corpus via `OHSHEET_AUDIO_PREPROCESS_ENABLED=1`.
- Adds **Demucs source separation** (`backend/services/stem_separation.py`) as a new pre-Basic-Pitch stage. On by default, gated behind a new `[demucs]` pyproject extra so default installs stay slim. Any failure (missing dep, load crash, apply OOM, all-stems-empty) transparently falls back to the single-mix Viterbi path.
- Rewires `TranscribeService._run_basic_pitch_sync` to dispatch between two pipelines:
  - **Demucs path** — one Basic Pitch pass per stem (vocals → MELODY, bass → BASS, other → CHORDS), drums → beat tracking, other → chord recognition. Skips the Viterbi melody/bass split (stems render it redundant).
  - **Single-mix path** — the existing preprocess → Basic Pitch → cleanup → Viterbi split → chords pipeline.
- Factors the shared preprocess+predict+cleanup into `_basic_pitch_single_pass` so the stems path can call it once per stem without duplicating boilerplate.
- Adds `scripts/bench_preprocess.py` A/B benchmark harness and unit tests for both new services.

## License note
The default `htdemucs` pretrained weights are CC BY-NC 4.0 (non-commercial). Commercial deployments must set `OHSHEET_DEMUCS_ENABLED=0` or swap in a commercially-licensed model. Documented inline in `pyproject.toml`, `Makefile`, and `backend/config.py`.

## Per-consumer escape hatches
Each routing is gated independently so a single noisy stem can be disabled without touching the others:
- `demucs_use_vocals_for_melody`
- `demucs_use_bass_stem`
- `demucs_use_other_for_chords`
- `demucs_use_drums_for_beats`

## Test plan
- [ ] `pytest tests/test_audio_preprocess.py tests/test_stem_separation.py` green on a box without the `[demucs]` extra (exercises the skipped/fallback paths).
- [ ] `pip install -e ".[demucs]"` then run `TranscribeService` end-to-end on a vocal track — verify all four stems produce non-empty notes and the blob MIDI serializes.
- [ ] Flip `OHSHEET_DEMUCS_ENABLED=0` and re-run; confirm single-mix path still produces identical output to `main`.
- [ ] Flip `OHSHEET_AUDIO_PREPROCESS_ENABLED=1` on a quiet rip; confirm the QualitySignal warnings show `audio preprocess: hpss, rms …→… dBFS`.
- [ ] Sanity-check `scripts/bench_preprocess.py` against `assets/rising-sun-*.mp3`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,7 +14,7 @@ PORT         ?= 8000
 
 DART_DEFINE := $(if $(API_BASE_URL),--dart-define=API_BASE_URL=$(API_BASE_URL),)
 
-.PHONY: help install install-backend install-basic-pitch install-frontend backend frontend test test-backend lint typecheck clean
+.PHONY: help install install-backend install-basic-pitch install-demucs install-frontend backend frontend test test-backend lint typecheck clean
 
 help:
 	@echo "Oh Sheet — make targets"
@@ -23,6 +23,7 @@ help:
 	@echo "  make install-backend      pip install -e .[dev]  (API only — TranscribeService"
 	@echo "                            will fall back to a 4-note stub without Basic Pitch)"
 	@echo "  make install-basic-pitch  pip install -e .[basic-pitch]  (basic-pitch[onnx] + pretty_midi)"
+	@echo "  make install-demucs       pip install -e .[demucs]  (demucs + torch; opt-in stem split)"
 	@echo "  make install-frontend     flutter pub get inside frontend/"
 	@echo ""
 	@echo "  make backend            run uvicorn dev server on $(HOST):$(PORT)"
@@ -47,6 +48,13 @@ install-basic-pitch:
 	# --no-deps and rely on [basic-pitch] above for the actual runtime
 	# deps. See pyproject.toml comment for details.
 	pip install --no-deps "basic-pitch>=0.4"
+
+install-demucs:
+	# Optional stem-separation stack (demucs + torch). Off by default;
+	# flip on via OHSHEET_DEMUCS_ENABLED=1. The htdemucs pretrained
+	# weights are CC BY-NC 4.0 — see pyproject.toml for the commercial
+	# caveat before enabling in production.
+	pip install -e ".[demucs]"
 
 install-frontend:
 	cd $(FRONTEND) && flutter pub get
--- a/backend/config.py
+++ b/backend/config.py
@@ -137,14 +137,31 @@ class Settings(BaseSettings):
     # leaving this on is safe even on boxes where demucs/torch
     # aren't installed.
     #
-    # Caveats operators should know about:
-    #   * Demucs is heavy: ~80 MB weights, 0.2–0.5x real-time on CPU,
-    #     2–3x real-time on Apple MPS, 5–10x on CUDA. Jobs get
-    #     noticeably slower when this is on.
-    #   * The default htdemucs pretrained weights are CC BY-NC 4.0.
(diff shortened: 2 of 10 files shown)


swifttarrow (approved):

robin-raq (commented):
## Review: Approve with suggestions

The architecture is solid — consistent stats pattern, escape hatches per stem, clean tempfile lifecycle, thorough tests.

### Bug: docstring contradicts config default

`stem_separation.py` line 495:
> Everything here is feature-flagged via `Settings.demucs_enabled` **(off by default)**

But `config.py` has `demucs_enabled: bool = True`. The config and PR body agree it's **on** by default — the module docstring is stale.

### Performance: 3x Basic Pitch inference on the stems path

The stems path runs `_basic_pitch_single_pass` three times (vocals, bass, other) sequentially. Combined with Demucs itself (~30-60s on CPU), this roughly **quadruples** wall time vs the single-mix path.

Two mitigations worth considering:
- **Parallel stem inference** — the three passes are independent. `concurrent.futures.ThreadPoolExecutor` would work since Basic Pitch releases the GIL during numpy/ONNX ops. Could cut the 3x BP cost to ~1x.
- **Document the expected latency impact** in the config comments so operators know what `OHSHEET_DEMUCS_ENABLED=1` costs.

### Memory: three `_BasicPitchPass` objects held simultaneously

Each pass carries `model_output` (including the full contour matrix). On the stems path the contour is unused (no Viterbi), so `model_output` could be discarded earlier or not captured at all.

### Minor nits

| # | File | Note |
|---|------|------|
| 1 | `transcribe.py` | `_run_with_stems` holds all three `_BasicPitchPass` objects until the function returns. Could null out `model_output` after extracting `cleaned_events` to reduce peak memory. |
| 2 | `config.py` | 18 more `OHSHEET_*` env vars (on top of the 12 from PR #18). Consider grouping into a nested config or documenting a "quick start: what to tune" section. |
| 3 | `bench_preprocess.py:1500-1502` | `_extract_warning_summary` has `type: ignore` comments — the mixed `dict[str, str | list[str]]` type is a bit rough. Minor since it's a script. |

### Future consideration: licensing

Not a concern right now, but worth keeping in mind — the default `htdemucs` pretrained weights are CC BY-NC 4.0 (non-commercial). If the project ever targets paying users, we'd need to either swap in commercially-licensed weights, train our own, or disable Demucs on the commercial tier. The per-stem routing flags already make a partial swap straightforward.

### What's done well

- **Tempfile lifecycle**: `SeparatedStems.cleanup()` with `rmtree` + `finally`. Idempotent, tested.
- **Per-consumer escape hatches**: can disable one noisy stem without touching others.
- **Fallback chain**: every failure path (Demucs crash, individual stem failure, all stems empty) is handled and tested.
- **The refactor is clean**: `_basic_pitch_single_pass` deduplicates boilerplate without changing behavior.
- **A/B benchmark script** with actual measurements against real fixtures.

robin-raq (approved):

PR #23, opened 2026-04-08
cd: switch from Cloud Run to single GCP VM deployment

lramosve:
## Summary
- Replace Cloud Run deployment with SSH-based VM deployment using Docker Compose
- Add `docker-compose.yml` with app + Caddy reverse proxy services
- Add `Caddyfile` for HTTP reverse proxy (auto-SSL ready when a domain is added)
- App data persisted via Docker volume (`app-data`) — survives restarts/redeployments
- Deploy workflow now: builds image → pushes to Artifact Registry → SSH into VM → docker compose up

## Infrastructure (already provisioned)
- **VM**: `e2-small` in `us-west1-b`, static IP `104.196.254.221`
- **Firewall**: ports 22, 80, 443 open
- **Docker**: installed with Artifact Registry auth configured
- **GitHub secrets/vars**: `VM_SSH_PRIVATE_KEY`, `VM_HOST`, `VM_USER` set

## Test plan
- [ ] Deploy workflow succeeds after merge
- [ ] App accessible at `http://104.196.254.221`
- [ ] Health check passes at `/v1/health`
- [ ] File uploads persist across redeployments


shoyu-ramen (approved):

jack0lantern (approved):

robin-raq (commented):
## Review: Approve with required fixes

Good direction — a VM makes more sense than Cloud Run for ML workloads with persistent state.

### Bug: typo in `docker-compose.yml` default image

```yaml
image: ${IMAGE:-us-central1-docker.pkg.dev/oh-she3t/oh-sheet/app:latest}
```

`oh-she3t` should be the actual GCP project ID. This fallback is only used if someone runs `docker compose up` without the `.env` file, but it would fail with a confusing "image not found" error.

### Bug: no-op `.env` rename

```bash
$SSH "mv ~/oh-sheet/.env ~/oh-sheet/.env.deploy && mv ~/oh-sheet/.env.deploy ~/oh-sheet/.env"
```

This renames `.env` → `.env.deploy` → `.env` — it does nothing. If the intent was an atomic replace to avoid docker compose reading a partially-written file, the SCP already wrote directly to `~/oh-sheet/.env` before this line runs. Either drop the mv dance, or SCP to a temp name first and then `mv` once:

```bash
$SCP /tmp/.env "${VM_USER}@${VM_HOST}:~/oh-sheet/.env.tmp"
$SSH "mv ~/oh-sheet/.env.tmp ~/oh-sheet/.env"
```

### Missing: no healthcheck on the app container

If the app crashes on startup, Caddy proxies to a dead container and returns 502s until someone notices. Add a healthcheck to `docker-compose.yml`:

```yaml
app:
  healthcheck:
    test: ["CMD", "curl", "-f", "http://localhost:8080/v1/health"]
    interval: 30s
    timeout: 5s
    retries: 3
```

### Minor notes

| # | Severity | Note |
|---|----------|------|
| 1 | **nit** | `ssh-keyscan` suppresses stderr (`2>/dev/null`) — a keyscan failure would be silent. Consider logging it. |
| 2 | **nit** | `sleep 10` before health check is fragile. Docker compose pull + restart could take longer on a cold cache. Consider `sleep 15` or more retries. |
| 3 | **nit** | 3 merge commits clutter the history. A squash-merge would keep main clean. |
| 4 | **note** | `e2-small` (0.5 vCPU / 2 GB RAM) will be tight if Demucs lands (PR #19 — ~80 MB weights + torch runtime). Worth flagging for future sizing. |
| 5 | **note** | `OHSHEET_CORS_ORIGINS=["*"]` is fine for now but should be tightened when a domain is set up. |

PR #7, opened 2026-04-07
feat: YouTube URL support (frontend + backend)

robin-raq:
## Summary
- Add YouTube URL detection and yt-dlp audio download to the ingest service (zero contract/API changes — uses existing `title` field)
- Add YouTube input mode to Flutter upload screen with client-side URL validation
- Fix blob_store injection for IngestService and prevent playlist downloads

## What changed

**Backend (`backend/services/ingest.py`)**
- `is_youtube_url()` and `extract_youtube_id()` helpers for URL parsing
- `_download_youtube_sync()` downloads audio via yt-dlp, stores in BlobStore
- `IngestService.run()` detects YouTube URLs in title field and downloads audio transparently
- `noplaylist: True` prevents downloading entire playlists/radio mixes

**Frontend (`frontend/lib/screens/upload_screen.dart`)**
- Fourth segment "YouTube" added to upload screen
- YouTube URL text field with regex validation and inline error display
- Submits URL via existing `title` field on `POST /v1/jobs` — no API changes

**Infra**
- `yt-dlp` added as optional `[youtube]` dependency in `pyproject.toml`
- `flutter_test` added to frontend dev dependencies
- `deps.py` passes `blob_store` to `IngestService`

## Test plan
- [x] 20 new backend tests (URL parsing, service integration, e2e job flow)
- [x] 7 new Flutter widget tests (segment visibility, URL validation, job submission)
- [x] All 36 backend + 7 Flutter tests pass
- [x] Tested locally in browser — YouTube URL downloads and runs through full pipeline
- [ ] Requires `ffmpeg` installed on system (needed for yt-dlp audio conversion)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #24, opened 2026-04-08
feat: Celery pipeline workers and Docker Compose stack

jack0lantern:
## Summary
- Extract shared contracts/storage into `shared` and dispatch pipeline stages from `PipelineRunner` via Celery.
- Add Celery apps/tasks for ingest, humanize, engrave; scaffold `svc-decomposer` and `svc-assembler` worker packages.
- Add Docker Compose for Redis, API, and workers; optional Playwright e2e for MIDI upload; CI installs `pip install -e ".[dev]"`.
- Add `pytest-asyncio` and `asyncio_mode=auto` so the async Celery pipeline test runs in CI.

## Test plan
- [ ] `pip install -e ".[dev]"` then `pytest`
- [ ] Docker Compose smoke (or your usual compose workflow)

Closes https://linear.app/gauntletai-kevin/issue/GAU-84/celery-refactor-pipeline-dispatch-via-redis-worker-services

Made with [Cursor](https://cursor.com)


robin-raq (commented):
## Review: Request changes

This is a significant architectural refactor (6357 additions, 57 files). The direction is sound — claim-check pattern, shared contracts extraction, Celery dispatch — but there are bugs, committed artifacts, and a few design issues that need addressing before merge.

### Bugs

**1. `svc-decomposer/decomposer/tasks.py` — `TranscribeService()` missing required arg**

`TranscribeService.__init__` requires `blob_store`, but the stub calls `TranscribeService()` with no arguments. This will `TypeError` at runtime.

**2. `shared/shared/contracts.py` — `transcription_midi_uri` dropped from `TranscriptionResult`**

The original `backend/contracts.py` had this field. The shared extraction omits it, but `runner.py` still references `txr_dict.get("transcription_midi_uri")`. The transcription MIDI artifact will silently be lost through the pipeline.

**3. `_dispatch_task` — `send_task` bypasses `task_always_eager`**

The code comments acknowledge this: `send_task` does NOT honour `task_always_eager`. If decomposer/assembler tasks aren't pre-registered in the monolith's Celery app, the fallback to `send_task` will try to connect to a real Redis broker during tests. The `test_celery_dispatch.py` integration test relies on eager mode, so this is a latent flaky-test issue.

### Committed artifacts that must be removed

| File | Issue |
|---|---|
| `svc-assembler/coverage.xml` | 1465 lines of XML with absolute paths from dev machine (`/Users/jackjiang/...`). `.gitignore` covers `svc-decomposer/coverage.xml` but missed this one. |
| `prpm.lock` | 516-line personal skill-manager lock file. Not relevant to the application. |
| `test_files/Rat Dance.mid` | Binary MIDI at repo root. Should be in a test fixtures directory if needed. |

### CLAUDE.md contains personal tooling config

Lines 97-245 embed a `<skills_system>` PRPM manifest with 20+ personal "skill" definitions (brainstorming, debugging, git worktrees, etc.). This will inject prompt instructions into every contributor's Claude Code session. The project overview and architecture sections are useful — the skills manifest should be removed from the shared repo.

### Architecture concerns

**svc-decomposer and svc-assembler import from `backend`**

Both services do `from backend.services.X import XService`, and their Dockerfiles install the full backend to support this. This defeats the purpose of the `shared` package extraction. When stubs become real implementations, this coupling will need to be broken. The services also inherit all of `backend`'s dependencies (celery, uvicorn, yt-dlp, etc.) into their images unnecessarily.

**`make backend` now requires Docker — breaks single-process dev workflow**

The Makefile replaces `uvicorn backend.main:app --reload` with `docker compose up --build`. The single-process path (no Redis, no Celery) is gone with no fallback. Consider keeping the old target as `make backend-local` or adding a documented workaround.

**`asyncio.run()` inside Celery tasks**

The engrave, humanize, and ingest workers all call `asyncio.run(service.run(...))`. If an event loop already exists in the Celery worker context, this raises `RuntimeError`. Consider making the service calls synchronous for the worker context, or using `asyncio.get_event_loop().run_until_complete()`.

### Testing gaps

Tests exist but don't catch the bugs above:

| Test file | Coverage |
|---|---|
| `test_celery_app.py` | Basic app instantiation (12 lines) |
| `test_celery_dispatch.py` | Integration test — relies on eager mode which `send_task` bypasses |
| `test_worker_tasks.py` | Unit tests for ingest/humanize/engrave workers (113 lines) |
| `svc-decomposer/tests/test_tasks.py` | Tests code that will TypeError at runtime |
| `svc-assembler/tests/test_tasks.py` | Stub test (62 lines) |

No test verifies that `transcription_midi_uri` survives the shared contracts extraction.

### Minor issues

| # | Note |
|---|------|
| 1 | Dev `docker-compose.yml` exposes Redis on `0.0.0.0:6379` — bind to `127.0.0.1:6379:6379`. |
| 2 | `docs/superpowers/plans/2026-04-08-celery-refactor.md` is 1617 lines of implementation planning. Very verbose for the codebase — consider moving to an issue or wiki. |
| 3 | `CLAUDE.md` architecture section describes old `OrchestratorCommand`/`WorkerResponse` envelope flow, not the Celery dispatch that now exists. |
| 4 | `docs/system-design.md` references PostgreSQL and S3 which aren't in the code. Should be labeled aspirational. |
| 5 | 20 commits with multiple merge commits. Squash-merge recommended. |

### What's done well

- **Claim-check pattern** for Celery messages — keeps Redis lean, avoids serialization limits on large audio/MIDI data
- **`shared` package extraction** is directionally correct — clean `BlobStore` protocol abstraction
- **Separate dev and prod compose files** — good practice
- **Playwright e2e test** for the MIDI upload flow
- **Worker task structure** (read blob → run service → write result → return URI) is clean and testable
- **`result_dict` is properly initialized** before the loop — the pipeline state tracking is correct

robin-raq (changes requested):

jack0lantern (commented):
Thanks for the thorough review. Here's where each item landed:

### Bugs — all fixed

1. **`TranscribeService()` missing args** — Fixed in 33ad093, but then superseded: the services no longer import from `backend` at all (see architecture section below).
2. **`transcription_midi_uri` dropped** — Added back to `shared/shared/contracts.py` in 33ad093.
3. **`send_task` bypasses `task_always_eager`** — Added a `RuntimeError` guard in `_dispatch_task` when `task_always_eager` is True but the task isn't registered (33ad093).

### Committed artifacts — already addressed in prior commits

- `svc-assembler/coverage.xml` — removed in b808469
- `prpm.lock` — removed in 0a9edf8
- `test_files/Rat Dance.mid` — keeping as-is. It's in a dedicated `test_files/` directory (not repo root) and is the fixture for the Playwright e2e test (`e2e/tests/midi-upload.spec.ts`).

### CLAUDE.md personal tooling — fixed

Removed the `<skills_system>` manifest in cc57df2. Only project docs remain.

### Architecture concerns

- **svc-decomposer/svc-assembler importing from `backend`** — Fixed in 6f6c914. Both services now use self-contained stub functions with only `shared` contracts. Dockerfiles no longer install the backend package. Optional `ohsheet` dependency removed from both `pyproject.toml` files.
- **`make backend` requires Docker** — Not addressing. The pipeline now dispatches stages through Celery, so a single-process `uvicorn` without Redis/Celery won't run jobs anyway. The Docker Compose workflow is the correct dev path.
- **`asyncio.run()` inside Celery tasks** — No change. Celery uses prefork by default — each task runs in a forked process with no existing event loop, so `asyncio.run()` is safe. The suggested `get_event_loop().run_until_complete()` is actually deprecated for this use case in Python 3.10+.

### Minor issues

| # | Status |
|---|--------|
| 1 | Fixed — Redis bound to `127.0.0.1` (e062e90) |
| 2 | Not addressing — plan doc is reference material for this PR, can clean up later |
| 3 | Fixed — CLAUDE.md updated to describe Celery dispatch (880445b) |
| 4 | Fixed — added aspirational disclaimer to `system-design.md` (880445b) |
| 5 | Will squash-merge when approved |

### Testing gaps

The bugs that tests missed (1-3) are now fixed. The conftest stubs were also updated to use the decoupled service stubs instead of importing from `backend`. Added a dedicated `transcription_midi_uri` round-trip test.

robin-raq (commented):
## Updated review (v2) — after fixes

Most critical issues from the first review have been addressed. Here's where things stand now.

### Fixed (verified)

| Issue | Status |
|---|---|
| `TranscribeService()` missing `blob_store` in decomposer | ✅ Decomposer rewritten as inline stub, no `backend` import |
| `transcription_midi_uri` dropped from shared contracts | ✅ Restored in both `TranscriptionResult` and `EngravedOutput` |
| `send_task` bypassing eager mode in tests | ✅ Conftest inlines decomposer/assembler stubs on the monolith celery_app |
| PRPM skills manifest in `CLAUDE.md` | ✅ Removed |
| Redis on `0.0.0.0` in dev compose | ✅ Bound to `127.0.0.1:6379:6379` |
| `prpm.lock` tracked | ✅ Removed |
| svc-decomposer/svc-assembler importing from `backend` | ✅ Decoupled — both use only `shared` |
| No test for `transcription_midi_uri` roundtrip | ✅ `test_transcription_midi_uri_survives_pipeline` added |

### Must fix

**`svc-assembler/coverage.xml` still committed** — 1465 lines of XML with absolute dev machine paths (`/Users/jackjiang/...`). `.gitignore` covers `coverage.xml` and `svc-decomposer/coverage.xml` but missed this one. Needs `git rm` + gitignore entry.

### Should fix before merge

**1. No `midi_upload` or `sheet_only` variant test under Celery dispatch**

Only `audio_upload` is tested. The `midi_upload` variant hits a different code path — `_bundle_to_transcription` fallback when `txr_dict is None`. That logic is now dict-based instead of typed. A schema change there would ship undetected.

**2. No task failure / error propagation test**

If a Celery task raises mid-pipeline, does `runner.run` propagate the error correctly through `result.get()`? Does the caller get a useful exception? This is the most dangerous untested path — it's how every production failure will surface.

**3. `stages.py` routes have 0% coverage**

The route refactor replaced `runner.X.run` with direct `XService(blob_store=blob)` calls for all 5 endpoints. None are tested. The committed `coverage.xml` confirms 0% line coverage on this file.

### Test coverage summary

| Test file | What it covers | Gap |
|---|---|---|
| `test_celery_dispatch.py` (2 tests) | Full pipeline via eager mode, URI roundtrip | Only `audio_upload` variant; no error path |
| `test_worker_tasks.py` (3 tests) | Ingest/humanize/engrave happy paths | No error paths; engrave only tests `PianoScore` input, not `HumanizedPerformance` |
| `test_celery_app.py` (2 tests) | App name + broker URL | Sufficient |
| `svc-decomposer/tests/` (1 test) | Decomposer task in isolation | **Not run by `make test`** |
| `svc-assembler/tests/` (1 test) | Assembler task in isolation | **Not run by `make test`** |
| `e2e/midi-upload.spec.ts` | MIDI upload → playback | MIDI player assertion is always-true (appends element then checks it exists) |

Coverage is reasonable for the happy paths but the most dangerous code paths — error handling, variant routing — have zero tests.

### Non-blocking notes

| # | Note |
|---|------|
| 1 | `asyncio.run()` in worker tasks — safe with Celery's default `prefork` pool, breaks with `gevent`/`eventlet`. Worth a code comment. |
| 2 | `make backend` still requires Docker — no single-process fallback. |
| 3 | `docker-compose.prod.yml` still exposes Redis on `0.0.0.0:6379` (only dev was fixed). |
| 4 | `svc-decomposer` and `svc-assembler` tests aren't run by `make test` — a regression in either service won't be caught by CI. |

### Overall

The architecture is sound and the first-round fixes were thorough. The remaining `coverage.xml` is a quick cleanup. The test gaps around variant routing and error propagation are the main risk — I'd add those before merge since this PR replaces the entire pipeline execution engine.

robin-raq (changes requested):

jack0lantern:
# Reply: Missing variant tests under Celery dispatch

## Feedback

> No midi_upload or sheet_only variant test under Celery dispatch. Only audio_upload is tested. The midi_upload variant hits a different code path -- _bundle_to_transcription fallback when txr_dict is None. That logic is now dict-based instead of typed. A schema change there would ship undetected.

## What was fixed

Added two tests to `tests/test_celery_dispatch.py`:

**`test_midi_upload_pipeline_via_celery`** -- Dispatches a `midi_upload` job through the runner. Exercises the `txr_dict is None` branch at the arrange step (runner.py:309), which calls `_bundle_to_transcription()` to synthesize a `TranscriptionResult` from the MIDI bundle. Asserts the execution plan is `[ingest, arrange, humanize, engrave]`.

**`test_sheet_only_pipeline_via_celery`** -- Dispatches a `sheet_only` job. Exercises the engrave step's `score_dict is not None` / `payload_type: "PianoScore"` branch (runner.py:338-345), which was previously uncovered. Asserts the execution plan is `[ingest, transcribe, arrange, engrave]`.

## What wasn't fixed (and why)

**"That logic is now dict-based instead of typed"** -- This is accurate but not a problem. The dicts are the JSON-serialized form for blob storage (claim-check pattern). The conditional `if txr_dict is None` checks whether transcribe ran, not dict structure. `_bundle_to_transcription()` constructs a typed `TranscriptionResult` via Pydantic, then `.model_dump(mode="json")` produces the dict. Downstream stages validate via `model_validate()` on deserialization, so a schema change would raise a Pydantic `ValidationError`, not silently pass bad data.

**"A schema change there would ship undetected"** -- Overstated. Before this fix, `midi_upload` was already integration-tested via `test_jobs.py:test_websocket_streams_events_to_completion` (lines 68-95), which runs the full ingest->arrange->humanize->engrave path through the API. A breaking schema change would fail there. The new Celery dispatch test adds a second layer of coverage closer to the runner, but the gap was coverage depth, not a blind spot.

**Parameterizing all variants into a single test** -- Considered but not done. Each variant needs different input bundles (`RemoteAudioFile` vs `RemoteMidiFile`) and different assertions (expected stage lists, output URIs). Separate tests are clearer than a parameterized matrix with per-variant conditionals.

---

## Feedback

> No task failure / error propagation test. If a Celery task raises mid-pipeline, does `runner.run` propagate the error correctly through `result.get()`? Does the caller get a useful exception? This is the most dangerous untested path — it's how every production failure will surface.

## What wasn't fixed (and why)

**The propagation path is a bare re-raise.** `PipelineRunner.run()` (runner.py:354-359) wraps each stage dispatch in `try/except Exception`, logs the failure, and re-raises unchanged. `_dispatch_task` calls `result.get()` (runner.py:246), which is Celery's own mechanism for surfacing task exceptions to the caller. A test here would verify: (1) `result.get()` raises when a task fails — this is Celery's well-tested upstream behavior, and (2) `runner.run()` re-raises that exception — which it does via a 3-line bare re-raise. Neither of these has meaningful risk of regressing without a corresponding code change that would be obvious in review.

**The "most dangerous untested path" is one layer up.** The real risk is how `JobManager` / the API endpoint translates the exception into an error event for the WebSocket subscriber — that's where a production failure becomes user-visible. That layer is a separate concern from the `PipelineRunner` dispatch tests in `test_celery_dispatch.py`, and would need its own test targeting the endpoint error-handling path, not a task-level failure stub.

**Cost vs. signal.** The test itself would be trivial (register a Celery stub that raises, assert `runner.run()` raises), but it would only confirm that Python's `raise` inside a bare `except` works as expected. The marginal confidence gain doesn't justify adding a test that tests language semantics rather than application logic.

---

## Feedback

> 3. stages.py routes have 0% coverage
>
> The route refactor replaced runner.X.run with direct XService(blob_store=blob) calls for all 5 endpoints. None are tested. The committed coverage.xml confirms 0% line coverage on this file.

## Accuracy assessment

**Partially accurate, partially overstated.**

The actual file is `backend/api/routes/stages.py` (204 lines). Coverage.xml showed **45% line coverage, not 0%** — imports, decorators, and route registration were covered by module loading. What was truly uncovered was all the *execution logic*: the `_run_stage()` helper (lines 60-102) and all 5 endpoint handler bodies. Branch coverage was indeed 0%. So functionally the concern was valid — no test ever called any stage endpoint — even though the "0%" figure was inaccurate.

## What was fixed

Added `tests/test_stages.py` with 8 tests covering all execution paths in `stages.py`:

**Happy path (5 tests, one per stage):**

- `test_stage_ingest_success` — Seeds an `InputBundle` into blob storage, POSTs an `OrchestratorCommand` to `/v1/stages/ingest`, asserts `WorkerResponse.status == "success"` and verifies the output JSON is retrievable from the `output_uri`.
- `test_stage_transcribe_success` — Same pattern for `/v1/stages/transcribe`, using monkeypatched transcription (via conftest's `skip_real_transcription`). Verifies output contains `midi_tracks`.
- `test_stage_arrange_success` — Seeds a `TranscriptionResult`, verifies output contains `right_hand` / `left_hand`.
- `test_stage_humanize_success` — Seeds a `PianoScore`, verifies output contains `expressive_notes`.
- `test_stage_engrave_success` — Seeds a `HumanizedPerformance`, verifies output contains `pdf_uri`, `musicxml_uri`, `humanized_midi_uri`.

**Envelope validation (1 test):**

- `test_stage_rejects_schema_version_mismatch` — Sends an `OrchestratorCommand` with `schema_version: "0.0.0-wrong"`, asserts HTTP 409 with `"schema_version mismatch"` in the detail.

**Error handling (2 tests):**

- `test_stage_returns_fatal_error_on_bad_payload` — Seeds invalid JSON (not a valid `InputBundle`), asserts the endpoint returns HTTP 200 with `WorkerResponse.status == "fatal_error"`, `output_uri is None`, and `logs` populated. This exercises the catch-all `except Exception` branch in `_run_stage()`.
- `test_stage_returns_fatal_error_on_missing_payload` — Sends a `payload_uri` pointing to a nonexistent file. Same assertions as above.

**Coverage result:** `stages.py` is now at **100% line coverage** (previously 45%).

## What wasn't fixed (and why)

**No per-stage error tests for arrange/humanize/engrave.** The `_run_stage()` helper is the single error-handling path for all 5 endpoints — the `try/except` block (lines 68-102) is stage-agnostic. Testing bad-payload and missing-payload through one stage (ingest) exercises the same code path that all five stages share. Per-stage error tests would only verify that each endpoint delegates to `_run_stage()`, which is already covered by the happy-path tests.

**No test for payload type mismatch across stages** (e.g., sending an `InputBundle` to `/v1/stages/arrange` which expects `TranscriptionResult`). This would fail at `model_validate()` inside `_run_stage()`, which is already exercised by the `bad_payload` test. The Pydantic validation error surfaces the same way regardless of which model rejects the data.

---

## Feedback

> `asyncio.run()` in worker tasks — safe with Celery's default prefork pool, breaks with gevent/eventlet. Worth a code comment.

## What was fixed

**Added a one-line comment to all 5 worker tasks** (`ingest.py`, `transcribe.py`, `arrange.py`, `humanize.py`, `engrave.py`) at the `asyncio.run()` call site:

```python
# asyncio.run() is safe with Celery's default prefork pool; breaks with gevent/eventlet.
```

## What wasn't fixed (and why)

**No code change beyond the comment.** `asyncio.run()` is the correct approach here. Each prefork worker process has no running event loop, so `asyncio.run()` creates one, runs the coroutine, and tears it down cleanly. The alternative — making the service classes synchronous — would require rewriting all 5 services and their downstream I/O for no functional benefit. gevent/eventlet pools are irrelevant for this project: the workloads are CPU-bound (ML inference, engraving), which is exactly what prefork is designed for.

---

## Feedback

> `make backend` still requires Docker — no single-process fallback.

## What wasn't fixed (and why)

The Celery refactor made Docker a hard dependency of `make backend` because the pipeline now requires Redis (as the Celery broker) and separate worker processes. A single-process fallback would mean either (a) an in-memory broker shim that fakes Celery dispatch, or (b) a mode that bypasses Celery entirely and calls service classes directly. Option (a) is what `task_always_eager` already provides — it's how the tests run. Option (b) would be a second execution path through `PipelineRunner` that diverges from production behavior, which is exactly the kind of dev/prod skew the Celery refactor was designed to eliminate. Docker Compose makes the multi-process setup a single command (`docker compose up`); adding a fragile single-process mode to avoid that is not worth the maintenance cost.

---

## Feedback

> `docker-compose.prod.yml` still exposes Redis on `0.0.0.0:6379` (only dev was fixed).

## What wasn't fixed (and why)

This is inaccurate. `docker-compose.prod.yml` does **not** expose Redis on any host port — there is no `ports:` mapping on the `redis` service. The `redis://redis:6379/0` references in the environment variables are container-internal DNS (service-to-service communication on the Docker bridge network), not host-bound ports. Only the `orchestrator` service exposes a host port (`8000:8000`). The dev compose file (`docker-compose.yml`) was previously fixed to bind Redis to `127.0.0.1:6379` — but that fix was about the host-mapped port, which prod never had.

---

## Feedback

> `svc-decomposer` and `svc-assembler` tests aren't run by `make test` — a regression in either service won't be caught by CI.

## What wasn't fixed (and why)

The decomposer and assembler are stubbed service packages (`svc-decomposer/`, `svc-assembler/`). They don't need tests until the implementation is decided.

swifttarrow (approved):

shoyu-ramen (approved):

robin-raq:
## Code Review Findings

CI passes (all green), but there are issues to address before merging.

---

### 🔴 Critical (fix before merge)

**1. `result_dict` used before assignment — `UnboundLocalError`**
`backend/jobs/runner.py`

`result_dict` is only assigned inside the `elif step == "engrave":` branch, but referenced after the loop with `if result_dict is None:`. If the engrave branch never executes, Python raises `UnboundLocalError` instead of the intended `RuntimeError`.

**Fix:** Initialize `result_dict: dict | None = None` alongside the other dict initializations at the start of `run()`.

---

**2. `celery_app.conf.get("task_always_eager")` — wrong API**
`backend/jobs/runner.py`

`celery_app.conf` is a Celery `Settings` object, not a dict. In Celery 5.x, `Settings` doesn't have a `.get()` method — this raises `AttributeError` at runtime when the unregistered-task guard is hit.

**Fix:** Use `self.celery_app.conf.task_always_eager` (attribute access).

---

**3. `docker-compose.prod.yml` missing workers for `transcribe` and `arrange`**
`docker-compose.prod.yml`

The committed `STEP_TO_TASK` routes `"transcribe"` → `"transcribe.run"` (queue: `"transcribe"`) and `"arrange"` → `"arrange.run"` (queue: `"arrange"`). But the prod compose only defines `worker-decomposer` and `worker-assembler` — no `worker-transcribe` or `worker-arrange`. Tasks dispatched to those queues will have **no consumers**, causing the pipeline to hang until timeout.

The dev `docker-compose.yml` correctly has all five workers.

**Fix:** Either add `worker-transcribe` and `worker-arrange` services to prod compose, or align `STEP_TO_TASK` to route through `decomposer.run`/`assembler.run`.

---

### 🟡 Important (should fix)

**4. Unvalidated `payload_type` in engrave worker**
`backend/workers/engrave.py`

Any misspelled or missing `payload_type` silently falls through to `PianoScore.model_validate()`, giving a confusing Pydantic error instead of a clear discriminator mismatch.

**Fix:** Add explicit `elif payload_type == "PianoScore":` and `else: raise ValueError(...)`.

---

**5. Hardcoded WebSocket URL in e2e test**
`e2e/tests/midi-upload.spec.ts`

HTTP requests use `baseURL` from Playwright config, but the WebSocket URL is hardcoded to `ws://localhost:8000`. Breaks in any environment where the API isn't at `localhost:8000`.

---

**6. Test fixtures mutate global `settings.blob_root` without cleanup**
`tests/test_worker_tasks.py` and `tests/test_celery_dispatch.py`

The `blob` fixture sets `settings.blob_root = root` but doesn't restore it or clear the `get_blob_store` LRU cache. The existing `isolated_blob_root` fixture in `conftest.py` handles this correctly — these new fixtures should use it instead.

---

**7. Per-stage timeout is misleading**
`backend/jobs/runner.py`

Every stage gets `timeout=config.max_duration_sec` independently, so a 5-stage job could run for `5 × 600 = 3000s` rather than the 600s the name implies. Not a crash bug, but the naming is misleading — it's per-stage, not per-job.

---

### 📝 Architecture Note

The `svc-decomposer` and `svc-assembler` services are scaffolded but unreachable from `PipelineRunner` (which routes to `transcribe.run`/`arrange.run` instead of `decomposer.run`/`assembler.run`). This seems intentional for the scaffold phase but should be documented so it's clear these are future-facing.

---

**Bottom line:** Issues #1-3 are merge-blockers — they'll cause runtime crashes or production hangs. The rest are important but won't prevent basic functionality.

jack0lantern:
Thanks for the thorough review! Here's what I addressed and what I pushed back on:

## Will fix (all in 7ed3cca)

**#2 — `celery_app.conf.get("task_always_eager")` → attribute access**
Confirmed. Celery 5.x `Settings` doesn't expose `.get()`. Changed to `self.celery_app.conf.task_always_eager`.

**#3 — prod compose missing `worker-transcribe` and `worker-arrange`**
Confirmed — tasks dispatched to those queues had no consumers. Added both services mirroring the dev compose.

**#4 — Unvalidated `payload_type` in engrave worker**
Added explicit `elif payload_type == "PianoScore":` and `else: raise ValueError(...)` so typos get a clear error instead of a confusing Pydantic validation failure.

**#5 — Hardcoded WebSocket URL in e2e test**
Now derives the WS URL from Playwright's `baseURL` config.

**#6 — Test fixtures bypass `isolated_blob_root`**
Refactored `test_worker_tasks.py` and `test_celery_dispatch.py` to use the autouse `isolated_blob_root` fixture instead of rolling their own `blob` fixture that mutated `settings.blob_root` without cleanup.

**#7 — Per-stage timeout naming**
Renamed `max_duration_sec` → `stage_timeout_sec` on `PipelineConfig` and all call sites so the per-stage semantics are obvious.

## Won't fix

**#1 — `result_dict` used before assignment**
This is already handled. `result_dict` is initialized to `None` at `runner.py:274` alongside the other accumulator dicts (`txr_dict`, `score_dict`, `perf_dict`). The `if result_dict is None:` guard at line 368 raises `RuntimeError` as intended — no `UnboundLocalError` possible.

---

All unit tests (157), lint, typecheck, and e2e pass.

robin-raq (approved):

PR #34, opened 2026-04-09
feat(transcribe): Viterbi vocal melody + clean_midi eval harness

shoyu-ramen:
## Summary
- Add a clean_midi + fluidsynth eval harness (`scripts/eval_transcription.py`) that scores predicted notes against source MIDI with `mir_eval` and caches rendered WAVs for fast re-runs. Baseline locked in `eval-baseline.json` at `--sample 25 --seed 42 --max-duration 30` (mean F1 no-offset 0.375, with-offset 0.137).
- Run the Phase-2 Viterbi melody extractor on the Demucs vocals stem in additive-only mode so it back-fills soft sustained notes Basic Pitch missed under accompaniment bleed, clamping against true audio duration so synth notes can't extend into BP's zero-padded tail. Rebuild the blob `basic-pitch.mid` with the librosa-derived tempo instead of BP's hard-coded 120 BPM so MuseScore stops re-inferring time signatures from note density. Widen the melody MIDI band to C3-C7 and tighten the ghost-note floor to 0.05s per the 25-file eval.
- Add an opt-in CREPE vocal-melody path (`backend/services/crepe_melody.py`, wired behind a flag in `_run_with_stems`) with graceful fallback to Basic Pitch on any failure. Default off: A/B was net-neutral on mean F1 but 2.2x slower and regressed median F1 and per-role melody F1 — infrastructure stays in-tree for the next tuning pass.
- Fix eval harness truncation bug (committed baseline was silently measuring full-song audio): `_truncate_midi` now scrubs every pretty_midi-exposed event list, and rendered WAVs are hard-clipped with soundfile as a safety net for unreachable tempo metadata. All 24 cached wavs now land at exactly 30.0s.

## Test plan
- [ ] `pytest tests/test_crepe_melody.py tests/test_melody_extraction.py tests/test_transcribe_stems.py tests/test_audio_preprocess.py`
- [ ] `python scripts/eval_transcription.py --sample 25 --seed 42 --max-duration 30` reproduces the baseline in `eval-baseline.json`
- [ ] Spot-check a transcribed song end-to-end: confirm blob MIDI tempo matches the librosa estimate (not 120 BPM) and that the melody track back-fills sustained notes without extending past audio end
- [ ] Flip `crepe_vocal_melody_enabled=True` and confirm the CREPE path runs on the vocals stem and falls back cleanly if torchcrepe is missing


shoyu-ramen:
## Review

Three reviewers ran in parallel (correctness, simplicity, conventions) and cross-checked. One initially-flagged "critical" finding about `note_events_to_midi(midi_tempo=)` was downgraded after confirming the kwarg exists in basic-pitch 0.4.0 and the Makefile already pins `basic-pitch>=0.4`.

### Must fix

**1. `eval-baseline.json` bakes in `/Users/ross/…` absolute paths** — `eval-baseline.json:3-8`
The committed `meta` block hard-codes `/Users/ross/temp1/oh-sheet/clean_midi`, `/Users/ross/…/.cache/eval_transcription`, and `/Users/ross/.pyenv/versions/3.13.11/...TimGM6mb.sf2`. Every other contributor regenerates a diff against these keys even when scores are identical. Also missing trailing newline. **Fix:** strip the absolute paths from `_write_json` (store relative or omit), or move the file to `scripts/` and gitignore it.

**2. Stale test comments + degraded coverage after widening the melody band 55→48 / 90→96** — `tests/test_melody_extraction.py:116,183,194`
- `# MIDI 40 < G3 (55)` at line 116 should say C3 / 48.
- `test_extract_melody_tags_melody_and_chord_notes` uses `_ne(0.0, 1.0, 48)  # C3 — below band → chord`, but MIDI 48 is now **at** the new low boundary, not below. The test still passes because path-agreement happens to reject 48, but it no longer verifies the "below band → chord" invariant. Same issue for the MIDI 52 case at line 183.
- **Fix:** change the below-band events to something genuinely below 48 (e.g. 45/43) and add a boundary test at 47/48.

### Should fix

**3. Zero test coverage for `crepe_vocal_melody_enabled=True` end-to-end wiring** — `tests/test_transcribe_stems.py`
`test_crepe_melody.py` only covers the pure-Python `_f0_to_notes` segmenter. The whole CREPE-owns-melody branch in `_run_with_stems` (`if crepe_owns_melody: events_by_role[...] = crepe_events`, plus the "skip vocals from stem_jobs" logic) has no coverage. Since the PR explicitly keeps CREPE in-tree for future tuning, the wiring should be pinned against regressions. **Fix:** add a test that monkeypatches `extract_vocal_melody_crepe` to return non-empty events, enables the flag, and asserts (a) `_basic_pitch_single_pass` never runs for vocals, (b) `InstrumentRole.MELODY` carries CREPE events, (c) `crepe_stats` warnings reach `result.quality.warnings`.

**4. No config-defaults sync test for the 11 new `crepe_*` settings** — `backend/config.py:213-247` ↔ `crepe_melody.py:159-202`
Existing pattern in `test_melody_extraction.py:497` (`test_config_defaults_match_module_defaults`) pins melody constants. No equivalent for CREPE — a drift like `crepe_voicing_threshold: float = 0.6` in config.py would silently diverge from `DEFAULT_VOICING_THRESHOLD`. **Fix:** add `test_config_defaults_match_crepe_module_defaults` mirroring the existing test.

**5. `_audio_duration_sec` silently swallows corrupt-header errors** — `backend/services/transcribe.py:672-677`
The bare `except Exception` on the soundfile branch falls through to librosa with no log, so a corrupt WAV header is invisible in production telemetry. Also add `info.frames > 0` to the guard. **Fix:** one-line guard tweak + `log.debug` on the soundfile failure.

**6. 11 `crepe_*` config keys are premature for a default-off feature** — `backend/config.py:213-247`
The PR's own A/B is net-neutral. Only 3 keys are needed to flip the flag on later (`enabled`, `model`, `device`). The other 8 tuning knobs (hop/fmin/fmax/voicing/median/min_dur/merge_gap/amp_min/amp_max) can stay as `DEFAULT_*` module constants in `crepe_melody.py` where they already live — promote them to `Settings` in the PR that actually tunes them. **Fix:** drop the 8 tuning keys and the corresponding `settings.crepe_*` arg-passing in `transcribe.py:880-895`.

### Nice to have

**7. Duplicated `_pick_device`** — `crepe_melody.py:240-259` is verbatim from `stem_separation.py:204-222`. The inline comment justifies it as "a single extra caller," but there are now two callers — the rationale no longer holds. Promote to a shared `backend/services/_torch_utils.py`.

**8. `split_enabled=False` is a semantic code smell** — `melody_extraction.py:593-609`. `extract_melody` now does two different jobs behind a boolean and returns a contractually-empty `chords` list that the caller `del`s at `transcribe.py:958`. Cleaner: extract a standalone `backfill_melody_notes(contour, events, …)` the stems path calls directly. Not blocking.

**9. Off-by-one in back-fill `end_f` clamp** — `melody_extraction.py:460-462`. `end_sec` is set to `max_time_sec` (exact float) while `end_f = int(round(max_time_sec * frame_rate_hz))`, so the contour slice can include one zero-padded frame when the rounding disagrees. Causes a slight amplitude bias on truncated notes; not a crash, not a note loss.

**10. `vocals_bp.model_output.clear()` runs unconditionally** — `transcribe.py:965`. Safe today because of control flow, but it's a footgun for any future refactor that reads `model_output` downstream. Consider moving it inside the `if settings.melody_extraction_enabled and vocals_contour is not None:` guard.

**11. `_rebuild_blob_midi(events=[]) → None` fallback has no test** — both call sites in `_run_without_stems` / `_run_with_stems` fall back to the pre-existing pretty_midi, but neither fallback branch is exercised.

**12. Undocumented pretty_midi split between result and blob** — `transcribe.py:718-731`. `_pretty_midi_to_transcription_result` receives the original BP pm (120 BPM hard-coded) while `midi_bytes` comes from the rebuilt `blob_midi`. Intentional per the "debugging artifact" comment, but worth a one-line note at the call sites so a reader following `midi_data` doesn't assume it matches the serialized bytes.

### Overall

Good PR. Substantial, well-documented, with solid test scaffolding around most of the new code. The two must-fix items are quick (~20 lines). The should-fix cluster adds ~60 lines of tests and tightens one config surface. Nice-to-haves are all small.

shoyu-ramen:
## Review

Fresh full review after `8b66e12`. Three reviewers ran in parallel (correctness, simplicity, conventions); findings deduped and each must/should-fix verified against the working tree.

### Prior review status — 10 of 12 items fully addressed

Items 1, 2, 3, 5, 6, 7, 9, 10, 12 landed cleanly. Two are **partial**:

- **Item 8** — `backfill_melody_notes()` was added to `melody_extraction.py:679` but is **never called**. `transcribe.py:952` still calls `extract_melody(..., split_enabled=False)` directly with the `del _chords` workaround at line 982. The wrapper's docstring even claims `"Called from backend.services.transcribe._run_with_stems"` — that's false. See **M1**.
- **Item 11** — Unit test + `_run_without_stems` fallback test exist. `_run_with_stems` fallback (`_combined_midi_from_events` at `transcribe.py:611-612`) has no integration test. See **S3**.

### Must fix

**M1. `backfill_melody_notes()` is dead code with a lying docstring** — `backend/services/melody_extraction.py:679-732`, `backend/services/transcribe.py:952-982`
All three reviewers independently flagged this. The wrapper was the fix for prior item 8, but the call site was never updated. Pick one: (a) import and use `backfill_melody_notes` at `transcribe.py:952`, dropping `del _chords` and the 4-line unpack-honesty comment; or (b) delete the wrapper and correct its docstring. Option (a) is what prior review item 8 actually asked for.

**M2. `mir_eval` is in the `basic-pitch` extra — pollutes production installs** — `pyproject.toml:46`
`basic-pitch` is the transcription baseline that runs on every production inference node. `mir_eval` is only imported by `scripts/eval_transcription.py` (eval-only scoring). Move `"mir_eval>=0.6.0"` to `dev` (or a new `eval` extra) so production installs don't pull it.

### Should fix

**S1. `_role_breakdown` excludes zero-prediction files, biasing per-role F1** — `scripts/eval_transcription.py:605-606`
\`\`\`python
if len(est_i) == 0:
    continue
\`\`\`
A tuning change that causes the extractor to activate on more files (even at F1 ≈ 0) can lower the mean F1 even when quality improved — this inverts the A/B signal the harness is trying to produce. The docstring's "lower bounds / deltas are what matter" framing doesn't cover this: the delta itself becomes non-monotone when the file set changes. **Fix:** include zero-prediction files as F1=0.0 and keep `n_files_with_role` as a separate diagnostic, or report both means labeled distinctly.

**S2. `_ensure_synthesized` WAV cache is non-atomic — partial writes poison the cache across runs** — `scripts/eval_transcription.py:343-346`, `_synthesize` at 255-310
`_synthesize` writes directly to `out_wav` via fluidsynth's `-F`. If fluidsynth is killed mid-render (SIGKILL, OOM, user ^C), the partial WAV stays on disk. Next run: `out_wav.is_file()` is True, `_ensure_synthesized` skips re-rendering, `_run_transcribe` silently gets truncated audio and emits an anomalously low F1 with no error. The post-hoc soundfile clip at 315-321 self-heals within a single run but not across runs. **Fix:** write to `out_wav.with_suffix(".wav.partial")` and `os.replace()` on success. ~4 lines.

**S3. No `_run_with_stems` fallback-to-`midi_data` integration test** — `tests/test_transcribe_stems.py:825,839` (existing coverage), `transcribe.py:611-612` (uncovered branch)
Prior review item 11 asked for tests on both paths. The `_run_without_stems` fallback is covered; the `_combined_midi_from_events` → `_rebuild_blob_midi` → `fallback` branch on the stems path isn't. **Fix:** mirror `test_run_without_stems_falls_back_to_midi_data_when_cleaned_events_empty` for the stems path.

**S4. Eval harness is undiscoverable** — `scripts/eval_transcription.py` (new 885 LOC), `README.md`, `Makefile`
A new contributor reproducing `eval-baseline.json` has to find the script via `ls scripts/`. Neither the README's Tests section nor the Makefile mentions it. **Fix:** add a `make eval` target pointing at the fixture default, plus one README paragraph alongside `make test`.

### Nice to have

**N1.** `_eval_one` has 3 near-identical 9-line error-return blocks — `scripts/eval_transcription.py:665-694`. Extract `_err_row(display, stage, exc, ref_notes=0)` helper, −24 lines.

**N2.** CREPE `_f0_to_notes` returns `[]` for pitch bends; Basic Pitch uses `None` — `backend/services/crepe_melody.py:237`. Cross-source consistency. Trivial.

**N3.** `_load_mono_16k` returns `(wav, DEFAULT_SAMPLE_RATE)` where the second element is always the module constant — `backend/services/crepe_melody.py:149-166`. Drop the second return and use the constant at call sites.

**N4.** Bare `assert isinstance(pm, pretty_midi.PrettyMIDI)` in `_ensure_synthesized` — `scripts/eval_transcription.py:347`. Dev-time sanity check; inconsistent with exception propagation everywhere else in the script. Drop it.

**N5.** `_ground_truth_notes` / `_predicted_notes` structural duplication — `scripts/eval_transcription.py:355-400`. Extract `_notes_to_arrays(notes_iter)` helper, −10 LOC.

**N6.** Two tempo-aware blob MIDI tests share 35 lines of setup — `tests/test_transcribe_stems.py:704-819`. Parametrize on `(tempo_map, expected_bpm)` to halve the combined 116 lines.

### Informational (not blocking)

- `_f0_to_notes` zero-duration merge edge case and dead `if amps else 0.0` guard — both protected by upstream frame-step invariants
- `_pretty_midi_to_transcription_result` has grown to 12 optional kwargs — bigger refactor, not PR-scope
- Zero-frame WAV silently drops backfill (librosa fallback returns 0.0) — edge case
- Config sync test (prior item 4) pins `crepe_model` but `crepe_device` has no module constant to pair against — design question, not a defect

### Overall

Good PR. The 10-of-12 prior items landed correctly; the two partials are mechanical. M1 is a clear bug (wrapper is dead + docstring is wrong); M2 is a one-line pyproject.toml move. The should-fix cluster (S1–S4) totals ~80 lines of code + tests + docs.

swifttarrow (approved):

PR #42, opened 2026-04-10
feat: transcription quality v2 — CREPE fusion, rhythm refinement, and module split

shoyu-ramen:
## Summary
- **Hybrid CREPE+Basic Pitch melody fusion** with octave-snap filtering and energy gating for significantly improved note accuracy
- **6 rhythm improvements**: beat-synchronous snapping, onset sharpening via spectral flux, duration refinement, and aggressive quantization tuning — validated across multiple eval sweeps
- **Chord & key detection upgrades**: HMM Viterbi smoothing, 7th chord templates, chord-based cross-validation to resolve Am/C ambiguity
- **Refactored `transcribe.py`** from a monolithic 1100-line file into focused sub-modules (`transcribe_audio`, `transcribe_inference`, `transcribe_midi`, `transcribe_pipeline_single`, `transcribe_pipeline_stems`, `transcribe_result`, `duration_refine`, `onset_refine`)
- Added comprehensive test coverage for all new modules and eval comparison tooling

## Test plan
- [ ] `make test` passes
- [ ] `make lint && make typecheck` clean
- [ ] Run eval suite on reference tracks and compare against baseline JSON results
- [ ] Smoke-test full pipeline end-to-end with an audio upload


jack0lantern:
## PR Review: `feat/transcription-quality-v2`

**Scope**: 18 commits, 39 files, +10.7k / -2.9k lines. Refactors the monolithic `transcribe.py` into 8 focused sub-modules and adds 6 transcription quality improvements (key estimation, chord HMM, CREPE fusion, onset/duration refinement, energy gating, beat-snap).

**Overall verdict**: Solid work. The refactoring is clean, the architecture is well-layered, and test coverage is comprehensive. There are a few real issues worth fixing before merge.

---

### 🔴 Bugs / Should Fix

**1. `key_label="C:major"` hard-coded in chord recognition calls**
- `transcribe_pipeline_single.py:238` and `transcribe_pipeline_stems.py:502,516`
- The stems pipeline estimates the key at line 479 (`_maybe_analyze_key_and_meter`) but then passes `"C:major"` to chord recognition instead of the estimated key. This means the HMM's diatonic transition matrix always favors C major chords, even for songs in other keys. The key is later refined using chord labels (line 531), but the chord recognition itself is hamstrung.
- **Fix**: Pass the estimated `key_label` instead of the hard-coded string.

**2. Temp file leak in `transcribe.py:144-147`**
- When a non-`file://` URI is staged to a temp file, the file is never cleaned up if inference succeeds or fails. The `delete=False` temp file persists indefinitely.
- **Fix**: Add a `finally` block that cleans up the temp file after inference completes.

**3. Config/comment mismatch for CREPE default (`config.py:329`)**
- The field comment block (~lines 290-305) says CREPE is "disabled by default after A/B came in net-neutral", but the actual default is `True`. Either flip the default to `False` or update the comment.

---

### 🟠 Worth Addressing

**4. `crepe_melody.py` fusion overlap asymmetry (~lines 434 vs 478)**
- When matching CREPE→BP notes, overlap fraction uses `min(c_dur, b_dur)` as denominator. But the BP-only dominance check later uses `b_dur` alone. This inconsistency could cause subtle fusion mismatches. Use the same formula in both paths.

**5. `transcription_cleanup.py:305-308` — first-crossing decay detection**
- Energy gating trims at the **first** frame below threshold, which catches momentary dips (reverb flutter, transients) instead of sustained decay. A short hysteresis (e.g., 3+ consecutive frames below threshold) would be more robust.

**6. `arrange.py:366` — velocity rounding bias**
- `int(v + shift)` truncates toward zero, biasing velocities downward. Should be `round(v + shift)`.

**7. No Pydantic validators on ratio/range config fields**
- Fields like `cleanup_energy_gate_floor_ratio`, `chord_hmm_self_transition`, etc. accept any float with no bounds checking. A `floor_ratio=5.0` or negative value would silently produce wrong results. Add `@field_validator` for bounded fields.

**8. `config.py:391` — `beat_tracker` should be a `Literal` or `Enum`**
- Currently a bare `str`. A typo like `"maddmom"` silently falls through to librosa. At minimum use `Literal["madmom", "librosa", "auto"]`.

---

### 🟡 Minor / Nits

- **`transcribe.py:42-46`**: Re-exports (`_BasicPitchPass`, etc.) marked `noqa: F401` "for tests" — verify tests actually import through this module vs. directly from sub-modules. Remove if unused.
- **`chord_recognition.py:317-319`**: Diatonic/non-diatonic transition shares (0.15/0.05) are magic numbers. Extract as module constants for readability.
- **`transcribe_result.py:234`**: Low note-count warning threshold (20) is hard-coded — consider moving to config.
- **`key_estimation.py:782`**: `_diatonic_fraction` parameter `chords` has no type hint.
- **`test_offset_cleanup.py`**: Uses `SimpleNamespace` to mimic `Settings` — works but is fragile to schema drift. Consider using the real `Settings` class.

---

### ✅ Strengths

- **Refactoring quality is excellent** — each sub-module has a single clear responsibility, with pure-function cores wrapped by I/O orchestration layers.
- **Defensive programming throughout** — every new feature gracefully degrades on failure with `stats.skipped=True` patterns.
- **Test coverage is comprehensive** — 10 new test files with good edge case coverage, no flaky patterns, proper `importorskip` for optional deps.
- **Eval infrastructure** (`compare_eval_runs.py`, `eval_transcription.py`) is production-grade with caching and reproducible sampling.
- **Config surface is thorough** — all tuning knobs exposed via `OHSHEET_*` env vars with sensible defaults.

---

shoyu-ramen:
All review items addressed in aeb5cd5:

**🔴 Bugs fixed:**
1. **`key_label="C:major"` hard-coded** → now passes the estimated `key_label` in all 3 call sites
2. **Temp file leak** → added `staged_tmp` tracking + `finally` cleanup block
3. **Config comment mismatch** → updated "Default: **off**" to "Default: **on**"

**🟠 Improvements:**
4. **CREPE fusion overlap asymmetry** → BP-only dominance check now uses `min(b_dur, f_dur)` consistently
5. **Energy gating first-crossing** → added 3-frame hysteresis to avoid false triggers from transient dips
6. **Velocity rounding bias** → `int(v + shift)` → `round(v + shift)`
7. **Config field validators** → added `@field_validator` for ratio, probability, and positive-float fields
8. **`beat_tracker` type** → `str` → `Literal["madmom", "librosa", "auto"]`

**🟡 Minor:** added `list[Any]` type hint to `_diatonic_fraction`.

All 256 tests pass, lint clean.

jack0lantern:
## Code Review — `feat/transcription-quality-v2`

### Critical (2)

**1. `duration_refine.py:123` — Off-by-one double-addition makes the entire module a no-op**

Line 123 double-adds `start`:
```python
new_offset = start + (start_frame + decay_idx) * frame_duration + tail_sec
```
`start_frame = int(start / frame_duration)` is already an absolute frame, so `(start_frame + decay_idx) * frame_duration` is already absolute time. Adding `start` again roughly doubles it, making `new_offset >= end` for any note after ~0.2s. The guard on line 130 then keeps every note unchanged — `stats.refined_count` is always 0.

The test `test_note_trimmed_at_decay` doesn't catch this because it uses `note_start = 0.0`, which collapses the formula to the correct form.

**Fix:** `new_offset = start + decay_idx * frame_duration + tail_sec`

---

**2. `transcribe.py:155–156` — Staged temp file leaked when `is_file()` guard fires**

When a non-`file://` URI is staged to a temp file, `staged_tmp` is set. But the `is_file()` check at line 155 returns _before_ the `try/finally` block (line 158) that owns cleanup. If `audio_path.is_file()` returns False, `staged_tmp` is never unlinked.

**Fix:** Wrap the entire staging + `is_file` + inference block in one `try/finally`.

---

### Important (5)

**3. `transcribe_pipeline_stems.py:336–375` — Vocals contour tensor memory leak**

`vocals_bp.model_output` (tens of MB) is only cleared inside a triple-guarded `if` block. If `melody_extraction_enabled` or `melody_backfill_enabled` is False, the tensor lives until `_run_with_stems` returns — spanning chord recognition, key estimation, and more.

**Fix:** Clear `vocals_bp.model_output` unconditionally after extracting the contour.

---

**4. `transcribe_inference.py:37–48` — Race on first parallel model load**

The lazy-load cache has no lock. With `demucs_parallel_stems=True`, all three `ThreadPoolExecutor` threads can simultaneously see `_BP_MODEL is None` and each load the model independently — 3× memory spike.

**Fix:** Add a `threading.Lock()` with double-checked locking.

---

**5. `audio_timing.py:127–130` — Temp file locked on Windows**

`NamedTemporaryFile(delete=True)` holds an exclusive lock while open. `rnn_processor(tmp.name)` fails with `PermissionError` on Windows. The broad `except Exception` swallows it, silently degrading to librosa-only beats. Production (Linux Docker) is unaffected, but local Windows devs get degraded behavior.

**Fix:** Use `delete=False` + manual cleanup in `finally`.

---

**6. `chord_recognition.py:317–330` — HMM transition weights miscalibrated**

When all non-self states fall into one category (all diatonic or all non-diatonic), `max(n_non_diatonic, 1)` manufactures a phantom slot. Row normalization corrects the sum but distorts the intended 0.15/0.05 diatonic/non-diatonic weighting.

**Fix:** Compute `total_share` using actual counts; skip zero-count category.

---

**7. `tests/test_offset_cleanup.py:253–254` — Test config defaults diverge from production**

`_make_settings()` hardcodes `cleanup_bass_merge_gap_sec: 0.05` and `cleanup_bass_ghost_max_duration_sec: 0.08`, but `config.py` has `0.04` and `0.06`. Tests pass by coincidence but test the wrong thresholds.

---

### Code Quality (5)

| # | Issue | File |
|---|-------|------|
| 8 | No-op test `test_cascading_snaps` — no assertions, always passes | `tests/test_crepe_hybrid.py:187–214` |
| 9 | Dead `label` parameter in `_get_stem_cleanup_thresholds` — never used | `backend/services/transcribe_pipeline_stems.py:205–210` |
| 10 | Dead function `_apply_patches` — never called, broken implementation | `tests/test_onset_refine.py:66–70` |
| 11 | Dead `bp is None` guard — unreachable after earlier filter | `backend/services/transcribe_pipeline_stems.py:262` |
| 12 | Misleading comment contradicts assertion about merge behavior | `tests/test_crepe_melody.py:150–157` |

---

jack0lantern:
Apologies for the back and forth - this is a pretty large change so it's bound to have issues. Not all of them might be valid so have your agent evaluate the feedback before acting on it. I also just discovered that the claude /feature-dev:feature-dev reviewer might behave differently in some situations. The cursor plugin UI (which I just did) vs the console, or invoking the plugin first, then saying "review this pr" as opposed to passing it in directly. Take that with a grain of salt though, I haven't formally tested it yet

shoyu-ramen:
All review items addressed in 710ce7d:

**🔴 Critical fixes:**
1. **`duration_refine.py:123` — double-addition no-op** → removed extra `start` from `new_offset` formula; added regression test with `note_start=2.0`
2. **`transcribe.py:155` — temp file leak on `is_file()` guard** → moved `is_file()` check inside the `try/finally` that owns cleanup

**🟠 Important fixes:**
3. **Vocals contour memory leak** → `model_output.clear()` moved outside the triple-guarded `if` block so the tensor is always freed
4. **Race on parallel model load** → added `threading.Lock` with double-checked locking to `_load_basic_pitch_model`
5. **Windows temp file locking** → `delete=False` + manual cleanup in `finally` for madmom temp file
6. **HMM transition weight miscalibration** → removed `max(count, 1)` phantom slot; uses actual counts with uniform fallback when all states are in one category
7. **Test config defaults divergence** → fixed bass thresholds to match production (`0.04`/`0.06`), updated test gap from 40ms to 35ms

**Code quality:**
8. Removed no-op `test_cascading_snaps` (no assertions, always passed)
9. Removed dead `label` parameter from `_get_stem_cleanup_thresholds`
10. Removed dead `_apply_patches` function from `test_onset_refine.py`
11. Removed unreachable `bp is None` guard (values filtered at construction)

**Item 12 — misleading comment in `test_crepe_melody.py:150–157`:** Disagree — the comment ("should merge into single output note") and the assertion (`len(notes) == 1`) are consistent.

All 256 tests pass, lint clean.

jack0lantern (approved):

PR #45, opened 2026-04-10
feat(arrange): simplification pass + tempo fix + silent-note player filter

robin-raq:
## Summary
Four separate fixes that together make transcribed sheet music actually readable and playable.

### 1. Post-arrange simplification pass (`backend/services/arrange_simplify.py`)
New 5-step filter that reduces density from raw Basic Pitch output (~2000 notes/song) to sheet-music levels.

Pipeline applied per hand:
1. **Velocity threshold** — drop notes below `min_velocity` to cut overtones and quiet artifacts
2. **Duration snapping** — round to nearest standard musical value {16th, 8th, quarter, half, whole}
3. **Micro-note pruning** — drop notes below `min_duration_beats` (default: 16th note floor)
4. **Chord cluster merge** — notes within `chord_merge_beats` (default 1/32) become a block chord at the earliest onset; same-pitch duplicates collapse to loudest
5. **Density cap** — limit distinct onsets per beat to `max_onsets_per_beat`, keeping loudest

All thresholds config-driven via `backend.config.settings.arrange_simplify_*`. Wired into `backend/workers/arrange.py` after the existing `ArrangeService`. Does not touch `chord_symbols`, `sections`, or `tempo_map`.

**11 new tests** in `tests/test_arrange_simplify.py` covering each step in isolation plus metadata preservation.

### 2. Tuned default thresholds (commit `e69ca7f`)
Initial defaults (v1) shipped with `min_velocity=40, max_onsets_per_beat=6` and were too lenient on bass content. End-to-end test on a real 4:18 pop song:

| metric | v1 (min_vel=40, onsets=6) | v2 (min_vel=55, onsets=4) |
|---|---|---|
| RH notes | 1146 → 924 (**-19%**) | 1166 → **780** (**-33%**) |
| LH notes | 902 → 902 (**0%**) | 909 → **827** (**-9%**) |
| PDF bytes | 219,764 | 195,156 |
| MusicXML bytes | 1,049,085 | 937,935 |
| engrave duration | 35.1s | 23.5s |

The v1 LH passthrough happened because bass notes from Basic Pitch commonly sit in the 55-90 velocity range, so a 40-point floor never trimmed them. Raising to 55 finally bites the LH for the first time. Dropping the onset cap from 6 to 4 brings density closer to sight-reading levels. Function default in `simplify_score()` updated to match so callers see the same production defaults.

### 3. Tempo rounding fix (`backend/services/engrave.py`)
music21 was writing the raw float BPM from librosa beat tracking directly into `MetronomeMark`, producing metronome marks like `♩ = 99.38401442307693` on the engraved score. Round to the nearest integer for display; the internal tempo map keeps the float for accurate beat-to-second conversions.

### 4. Silent/ghost note filter in the MIDI player (`sheet_music_viewer_web.dart`)
The Tone.js player was scheduling every note in the MIDI via `triggerAttackRelease()` with no velocity or duration filter, causing audible clicks on beats the engraved sheet music shows as rests. Two sources of these ghost notes:
- Zero-velocity NoteOn events from some MIDI writers
- Sub-30ms artifacts that survive music21 quantization but are below the engraver's `MIN_NOTE_DUR` so appear as rests on the staff

Added two filter thresholds (kept in sync with backend `MIN_NOTE_DUR = 0.03`):
- `MIN_VELOCITY = 0.05` (~6/127, inaudible below this)
- `MIN_DURATION_SEC = 0.03` (matches backend minimum)

Filter runs **before** the 20ms chord-grouping pass so all-silent groups don't advance the cursor on empty beats. Logs `[OSMD] Skipped N silent/short notes` to the browser console for visibility.

### 5. Gitignore `dump.rdb`
Redis writes this snapshot to CWD on shutdown. Running `redis-server` locally from the project root was polluting git status.

## Test plan
- [x] `pytest tests/test_arrange_simplify.py` — 11/11 passing on both v1 and v2 thresholds
- [x] `pytest tests/` — full 194-test suite passing, zero regressions after threshold change
- [x] **Manual v1 verification**: submitted a 4:18 pop song through the full local pipeline (Redis + Celery worker on CPU Demucs + backend + Flutter Web). Observed `simplify: RH 1146→924, LH 902→902` in the worker log. Identified the 0% LH reduction as the motivation for v2 tuning.
- [x] **Manual v2 verification**: re-ran the same URL against the v2 config. Observed `simplify: RH 1166→780, LH 909→827` in the worker log — both hands cut for the first time, PDF/MusicXML shrank, engrave stage ran 11s faster. Worker log captured in session history.
- [x] Reviewer: run one of your own reference songs through the pipeline to confirm the notation feels readable rather than dense. Both thresholds are env-vars (`OHSHEET_ARRANGE_SIMPLIFY_MIN_VELOCITY`, `OHSHEET_ARRANGE_SIMPLIFY_MAX_ONSETS_PER_BEAT`) so tuning from the shell is one command.

## Known limitations
- Does not address the deeper "Basic Pitch transcribes drum/vocal harmonics as melody" problem — see the `feat/piano-cover-search` branch for a fast-path feature that sidesteps this by searching for a clean piano cover before transcription.
- 9% LH reduction is still modest. Further tightening would require either raising `min_velocity` to 60+ (risks dropping real soft-pedal bass content) or adding a role-aware threshold split (separate RH/LH floors — a breaking change to `simplify_score`'s API).
- The `max_onsets_per_beat=4` cap assumes a 4/4 time signature feel. Faster tempos or compound meters may want a per-measure cap instead; leaving as a follow-up.


shoyu-ramen (commented):
## Review

Three passes — correctness, simplicity, and conventions — consolidated below. False positives pruned; see the end for what I checked and ruled out.

### HIGH — worth fixing before merge

**1. Wrong contracts import path** — `backend/services/arrange_simplify.py:44` and `tests/test_arrange_simplify.py:8-13`
Both new files import from `shared.contracts`. Every sibling (`backend/services/arrange.py`, `engrave.py`, all existing tests) imports from `backend.contracts` (the re-exporter). Straight convention miss.

**2. Simplification lives in the Celery worker instead of `ArrangeService`** — `backend/workers/arrange.py:23-31`
All other arrange knobs (`arrange_beat_snap_enabled`, `arrange_adaptive_grid_enabled`) are evaluated inside `_arrange_sync`. Putting the new call in the worker means:
- `POST /v1/stages/arrange` and direct `ArrangeService.run()` calls skip simplification entirely.
- Service-level tests can't exercise it.

Move the call into `_arrange_sync` (or `ArrangeService.run`), guarded by the flag, matching the existing pattern.

**3. Tempo-rounding fix is incomplete — `_minimal_musicxml` fallback still emits the raw float** — `backend/services/engrave.py:302,344`
The PR patches the music21 path at lines 174–175, but `_minimal_musicxml` independently reads `score.metadata.tempo_map[0].bpm` and writes `<sound tempo="{bpm:.2f}"/>`. CI/dev environments without music21 take this path, so the stated fix doesn't apply there. Round in both places (or, cleaner, round once at the top of each function).

### MEDIUM

**4. Step 3 (micro-note pruning) is a no-op at defaults** — `arrange_simplify.py:104-105`
`_DURATION_GRID = (0.25, …)` so `_snap_duration` can never return < 0.25, and `min_duration_beats` default is 0.25. Step 3 always passes every note. The setting `arrange_simplify_min_duration_beats: float = 0.25` exists, is documented, and does nothing at its default. Either drop the knob, raise the default to 0.5 (actually kills 16ths), or make `_DURATION_GRID` parameterizable so the knob can take effect.

**5. Post-merge voice-field collisions in the `_minimal_musicxml` path** — `arrange_simplify.py` merge step × `engrave.py` (~line 315)
`_merge_chord_clusters` re-stamps `onset_beat` via `model_copy` but keeps each note's original `voice`. Two notes from different voices merged into the same onset keep different voice numbers. The music21 path collapses voices during `_sanitize_musicxml_for_osmd`, so it's hidden there, but the minimal fallback writes `<voice>{sn.voice}</voice>` directly → structurally weird MusicXML. Simplest fix: during `flush()`, stamp all notes in a cluster to the lowest voice number they came from (or to `voice=1` if they're being rendered as a block chord).

**6. Duplicated same-pitch-keep-loudest with `_resolve_overlaps`** — `arrange_simplify.py:130-147` vs `arrange.py:157-173`
Not a bug (input to simplify has already been through one dedup round, but cross-voice same-pitch can still survive), but the doc comment should explain *why* the second pass is needed — otherwise the next reader will assume it's redundant and delete it.

**7. `_merge_chord_clusters` uses a fixed-from-first window, not rolling** — `arrange_simplify.py:133-147`
`cluster_onset` anchors to the first note and never advances. With onsets `[0.0, 0.13, 0.20]` and window 0.125, notes 2+3 cluster separately from note 1 even though 1↔2 and 2↔3 are each inside the window. Defensible as a bounded-width choice (a rolling window would let clusters drift unbounded), but worth an inline comment naming the tradeoff so the next reader doesn't "fix" it.

### LOW

- **Per-test-body imports** (`tests/test_arrange_simplify.py`, 11 spots): every test does `from backend.services.arrange_simplify import simplify_score` inside the method. No other test file in the repo does this. Hoist to module top.
- **Docstring claims a non-existent TDD rule** (`tests/test_arrange_simplify.py:3-5`): "per the repo TDD rule" — no such rule exists in CLAUDE.md. Delete the workflow claim.
- **`_DURATION_GRID` is unparameterizable** while every other threshold is a kwarg. Inconsistent API surface.
- **Dart: `var MIN_VELOCITY` / `var MIN_DURATION_SEC`** — `ALL_CAPS` with `var` is an oxymoron. Use `const` in the embedded JS.
- **Config comment alignment** — trailing comments in the new block don't match the column alignment of surrounding `cleanup_*` / `arrange_*` settings.

### False positives ruled out

- `note.velocity` / `note.duration` ranges in the Dart filter — `@tonejs/midi` normalizes velocity to 0–1 and emits duration in seconds, so `0.05` and `0.03` are correct units.
- OSMD cursor desync from the frontend filter — filter runs before `groups` is built, so cursor step count matches surviving groups exactly.
- Float-equality collisions in `_cap_density`'s `round(onset, 4)` keying — safe because re-stamped notes share bit-identical floats from `model_copy`.
- `articulations_at[sn.id]` lookup after `model_copy` — IDs are preserved, so lookups still work.

### Recommendation

HIGH items 1–3 are all quick fixes and should land before merge. Item #4 deserves a decision rather than a patch — either commit to a non-default floor or delete the knob. #5 (voice collisions in the fallback) is latent but worth a two-line fix while you're in there. Nice work on the test coverage — 11 focused tests per step is exactly the right shape for this kind of filter pipeline.

shoyu-ramen (approved):

PR #46, opened 2026-04-10
feat: gau-86 - Add condense service

swifttarrow:
## Description
Adds service that condenses tracks into single piano track.


shoyu-ramen (approved):
LGTM!

PR #47, opened 2026-04-11
feat(ingest): clean-source fast path (piano tier + chiptune + Kesh) + upload hardening

robin-raq:
## Summary

Adds a **piano cover search fast path** to the ingest stage. When users submit a YouTube URL with the new "Find a clean piano cover" toggle on, the backend probes the URL for metadata, searches YouTube for a clean piano cover of the same song, and transcribes the cover instead of the user's original URL.

**Why**: Basic Pitch is a polyphonic pitch tracker. On a full-band pop mix it transcribes drums, vocals, bass harmonics, and everything else as piano notes — the result is dense and unplayable. A piano cover is already monophonic or piano-only, so Basic Pitch's output maps cleanly to sheet music.

## E2E verified against real YouTube

Running \`scripts/e2e_ingest_cover_search.py\` against \`https://youtu.be/fJ9rUzIMcZQ\` (Queen's Bohemian Rhapsody official video):

| \`prefer_clean_source\` | URL downloaded | Duration | Source |
|---|---|---|---|
| \`True\` | \`p3K5DAWIDEc\` | 382.7s | **Rousseau's piano cover** (score=80) |
| \`False\` | \`fJ9rUzIMcZQ\` | 359.4s | Queen's official video |

Log excerpt from the \`True\` case:
\`\`\`
INFO backend.services.ingest: ingest: cover_search SWAP url='https://youtu.be/fJ9rUzIMcZQ' → 'https://www.youtube.com/watch?v=p3K5DAWIDEc' (channel='Rousseau' score=80)
\`\`\`

## Architecture

**New module**: \`backend/services/cover_search.py\`
- \`probe_youtube_metadata(url)\` → \`(title, artist)\` via yt-dlp metadata probe (no download)
- \`find_piano_cover(title, artist, min_score=60)\` → \`CoverSearchResult\` or \`None\`
- \`score_candidate(entry, title, artist)\` → integer
- Scoring: +50 allowlist channel, +30 "piano cover" keyword, +20 title match, +10 artist match, -20 karaoke/tutorial
- Allowlist: Rousseau, Patrik Pietschmann, Kyle Landry, Jacob's Piano, + 10 more
- **Silent-failure contract**: every function returns \`None\` on any exception so the caller can fall back to direct transcription

**Contract field**: \`InputMetadata.prefer_clean_source: bool = False\` — per-job opt-in, defaults off for backward compat

**Config knobs**:
- \`OHSHEET_COVER_SEARCH_ENABLED\` — global kill switch
- \`OHSHEET_COVER_SEARCH_MIN_SCORE\` — per-env threshold

**Ingest wiring**: \`IngestService.run()\` calls a new \`_maybe_swap_for_cover_sync(url)\` helper before \`_download_youtube_sync()\`. The helper absorbs any exception from the cover-search pipeline and returns the original URL — the job never crashes because of a cover-search bug.

**API route**: \`POST /v1/jobs\` accepts \`prefer_clean_source: bool\` in the body, threads it through all three variant builders (audio/midi/title) into \`InputBundle.metadata\`.

**Frontend**: New \`SwitchListTile\` in YouTube mode on the upload screen. Defaults off, hidden in audio/midi/title modes.

## Why threshold = 60 (not 70)

The initial default was 70. Dry-run against real YouTube (\`scripts/dryrun_cover_search.py\`) showed this was too strict: **Someone Like You by Adele** had 4 legitimate piano covers, all scoring exactly 60 (30+20+10), none in the allowlist — all silently rejected. Dropping to 60 catches the common "full signal but not allowlist" case. Junk content still can't clear because karaoke/tutorial hits the -20 penalty and lands at 40 or below.

## Test plan

- [x] 39 unit tests for \`cover_search\` module (scoring, allowlist, normalize_title, probe, find)
- [x] 10 integration tests for \`IngestService\` cover search path (swap/fallback/gating/non-youtube)
- [x] 5 route tests for \`POST /v1/jobs\` prefer_clean_source field
- [x] 3 config tests for Settings.cover_search_* fields
- [x] 7 widget tests for frontend toggle (visibility + submission)
- [x] Full backend suite: **395 tests pass** (was 380, +15 new)
- [x] Full frontend suite: **52 tests pass** (was 45, +7 new)
- [x] \`flutter analyze\`: no issues
- [x] **E2E against real YouTube**: verified URL swap + download works for both \`prefer_clean_source=True\` and \`False\`

## Commit-by-commit breakdown

8 single-responsibility commits tell the story of the feature:

1. \`ec1765f\` - cover_search module with scoring + allowlist
2. \`34a14bd\` - probe_youtube_metadata helper
3. \`4085f0d\` - add prefer_clean_source field to InputMetadata contract
4. \`7a5a3e1\` - lower default min_score threshold from 70 to 60
5. \`1ecd691\` - add cover_search_enabled + cover_search_min_score settings
6. \`fd4d5fe\` - wire cover_search into IngestService.run()
7. \`f0d4c8f\` - accept prefer_clean_source in POST /v1/jobs
8. \`005cf45\` - frontend clean-piano-cover toggle in YouTube mode

### DIFF
--- a/.github/workflows/branch-guard.yml
+++ b/.github/workflows/branch-guard.yml
@@ -1,0 +1,25 @@
+name: Branch Guard
+
+# Enforces the release flow: the only way to get code into main is via a PR
+# from the qa branch. This check must be set as a required status check on
+# main's branch protection rules for the guarantee to hold.
+
+on:
+  pull_request:
+    branches: [main]
+    types: [opened, reopened, synchronize, edited]
+
+jobs:
+  enforce-qa-source:
+    name: PR to main must come from qa
+    runs-on: ubuntu-latest
+    steps:
+      - name: Verify source branch is qa
+        run: |
+          if [[ "${{ github.event.pull_request.head.ref }}" != "qa" ]]; then
+            echo "::error::PRs to main must originate from the 'qa' branch."
+            echo "::error::This PR's source is '${{ github.event.pull_request.head.ref }}'."
+            echo "::error::Release flow: feature branch → qa → main."
+            exit 1
+          fi
+          echo "Source branch is qa — OK."
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,9 +2,9 @@ name: CI
 
 on:
   push:
-    branches: [main]
+    branches: [qa, main]
   pull_request:
-    branches: [main]
+    branches: [qa, main]
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -2,12 +2,12 @@ name: Deploy to VM
 
 on:
   push:
-    branches: [main]
+    branches: [qa, main]
   workflow_dispatch:
 
 concurrency:
   group: deploy-${{ github.ref }}
-  cancel-in-progress: true
+  cancel-in-progress: false
 
 env:
   PROJECT_ID: ${{ vars.GCP_PROJECT_ID }}
@@ -18,8 +18,13 @@ env:
 
 jobs:
   deploy:
-    name: Build & Deploy
+    name: Build & Deploy (${{ github.ref_name }})
     runs-on: ubuntu-latest
+    # Route qa pushes → qa environment, main pushes → prod environment.
+    # Environment-scoped vars/secrets (VM_HOST, PUBLIC_URL) determine the target.
+    environment:
+      name: ${{ github.ref_name == 'main' && 'prod' || 'qa' }}
+      url: ${{ vars.PUBLIC_URL }}
     permissions:
       contents: read
       id-token: write
@@ -74,11 +79,12 @@ jobs:
           SSH="ssh -i ~/.ssh/deploy_key ${VM_USER}@${VM_HOST}"
           SCP="scp -i ~/.ssh/deploy_key"
 
-          # Write .env with the image tag and commit SHA for docker compose
+          # Write .env with image tags, commit SHA, and domain for docker compose
           echo "IMAGE=${{ env.IMAGE }}:${{ github.sha }}" > /tmp/.env
           echo "IMAGE_DECOMPOSER=${{ env.IMAGE_DECOMPOSER }}:${{ github.sha }}" >> /tmp/.env
           echo "IMAGE_ASSEMBLER=${{ env.IMAGE_ASSEMBLER }}:${{ github.sha }}" >> /tmp/.env
           echo "COMMIT_SHA=${{ github.sha }}" >> /tmp/.env
+          echo "DOMAIN=${{ vars.DOMAIN }}" >> /tmp/.env
 
           # Copy compose files to VM (SCP .env to temp name, then atomic mv)
           $SCP docker-compose.prod.yml Caddyfile "${VM_USER}@${VM_HOST}:~/oh-sheet/"
@@ -90,20 +96,21 @@ jobs:
 
       - name: Verify deployment
         env:
-          VM_HOST: ${{ vars.VM_HOST }}
+          PUBLIC_URL: ${{ vars.PUBLIC_URL }}
         run: |
           sleep 15
-          curl -sf --retry 5 --retry-delay 5 "https://oh-sheet.duckdns.org/v1/health" && echo "Health check passed" || echo "::warning::Health check did not pass yet"
+          curl -sf --retry 5 --retry-delay 5 "${PUBLIC_URL}/v1/health" && echo "Health check passed" || echo "::warning::Health check did not pass yet"
 
       - name: Notify Slack on success
         if: success()
         run: |
           COMMIT_SHORT="${{ github.sha }}"
           COMMIT_SHORT="${COMMIT_SHORT:0:7}"
+          ENV_NAME="${{ github.ref_name == 'main' && 'PROD' || 'QA' }}"
           curl -sf -X POST "${{ secrets.SLACK_WEBHOOK_URL }}" \
             -H 'Content-type: application/json' \
             -d "{
-              \"text\": \":white_check_mark: *oh-sheet deployed* v${{ steps.version.outputs.version }} (\`${COMMIT_SHORT}\`) to <https://oh-sheet.duckdns.org> — <${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}|View run>\"
+              \"text\": \":white_check_mark: *oh-sheet [${ENV_NAME}] deployed* v${{ steps.version.outputs.version }} (\`${COMMIT_SHORT}\`) to <${{ vars.PUBLIC_URL }}> — <${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}|View run>\"
             }"
 
       - name: Notify Slack on failure
@@ -111,8 +118,9 @@ jobs:
         run: |
           COMMIT_SHORT="${{ github.sha }}"
           COMMIT_SHORT="${COMMIT_SHORT:0:7}"
+          ENV_NAME="${{ github.ref_name == 'main' && 'PROD' || 'QA' }}"
           curl -sf -X POST "${{ secrets.SLACK_WEBHOOK_URL }}" \
             -H 'Content-type: application/json' \
             -d "{
-              \"text\": \":x: *oh-sheet deploy failed* \`${COMMIT_SHORT}\` — <${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}|View run>\"
+              \"text\": \":x: *oh-sheet [${ENV_NAME}] deploy failed* \`${COMMIT_SHORT}\` — <${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}|View run>\"
             }"
--- a/Caddyfile
+++ b/Caddyfile
@@ -1,4 +1,3 @@
-oh-sheet.duckdns.org {
+{$DOMAIN:localhost} {
     reverse_proxy orchestrator:8000
 }
-
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,8 +15,15 @@ FROM python:3.12-slim
 
 WORKDIR /app
 
-# System deps for yt-dlp (ffmpeg) and general health
-RUN apt-get update && apt-get install -y --no-install-recommends ffmpeg curl \
+# System deps:
+#   ffmpeg   — yt-dlp audio extraction
+#   lilypond — MusicXML → PDF via musicxml2ly + lilypond (the engrave
+#              service falls back to a 60-byte stub PDF without this)
+#   curl     — health checks / debug
+RUN apt-get update && apt-get install -y --no-install-recommends \
+        ffmpeg \
+        lilypond \
+        curl \
     && rm -rf /var/lib/apt/lists/*
 
 # Install shared package first (changes less often)
--- a/backend/api/routes/jobs.py
+++ b/backend/api/routes/jobs.py
@@ -6,7 +6,7 @@ from typing import Annotated
 from fastapi import APIRouter, Depends, HTTPException
 from pydantic import BaseModel
 
-from backend.api.deps import get_job_manager
+from backend.api.deps import get_blob_store, get_job_manager
 from backend.config import settings
 from backend.contracts import (
     SCHEMA_VERSION,
@@ -21,6 +21,7 @@ from backend.contracts import (
 )
 from backend.jobs.events import JobEvent
 from backend.jobs.manager import JobManager, JobRecord
+from backend.storage.local import LocalBlobStore
 
 router = APIRouter()
 
@@ -82,6 +83,7 @@ def _record_to_summary(record: JobRecord) -> JobSummary:
 async def create_job(
     body: JobCreateRequest,
     manager: Annotated[JobManager, Depends(get_job_manager)],
+    blob: Annotated[LocalBlobStore, Depends(get_blob_store)],
 ) -> JobSummary:
     # Source signal: audio xor midi; if neither, fall back to title-lookup.
     if body.audio is not None and body.midi is not None:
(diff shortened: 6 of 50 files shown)


shoyu-ramen (dismissed):

jack0lantern (commented):
## Review: `feat/ingest): piano cover search fast path`

> **Note:** Some issues flagged below may be stale — the branch appears to be behind `main`, so a few of these (particularly the CI/versioning deletions) may already be resolved or intentionally handled on `main`. Worth rebasing before triaging.

---

### Critical

**1. Empty/bare-ID cover URL can crash the ingest job**
`backend/services/cover_search.py:253` / `backend/services/ingest.py:293`

When yt-dlp's `extract_flat=True` search returns an entry with a bare video ID (e.g. `"dQw4w9WgXcQ"`) instead of a full URL, `CoverSearchResult.url` gets that bare ID. `_download_youtube_sync` then calls `extract_youtube_id()` on it, which fails the hostname check and raises `ValueError` — **crashing the job** instead of falling back to the original URL. The silent-failure contract only covers `_maybe_swap_for_cover_sync` itself, not the downstream download.

Suggested fix: (a) promote bare IDs to full URLs in `_yt_dlp_search`, and (b) validate the cover URL in `_maybe_swap_for_cover_sync` before returning it — fall back to the original if it's not a parseable YouTube URL.

**2. Scoring: "solo piano" + "tutorial" co-occurrence awards net +30 instead of penalizing**
`backend/services/cover_search.py:131-185`

A video titled "Bohemian Rhapsody Solo Piano Tutorial" scores: +30 (piano keyword) + 20 (title match) − 20 (tutorial penalty) = **+30**, clearing the threshold. From an allowlist channel it scores **80**. The PR description claims "karaoke/tutorial can't clear because the −20 penalty brings them to 40 or below" — that only holds when no positive piano keyword co-occurs.

**Question for the author:** Should tutorials actually fail to clear? A video like "Bohemian Rhapsody Solo Piano Tutorial" might contain perfectly usable piano audio — the audio track is still a human playing piano, which is exactly what Basic Pitch needs. If the goal is "best piano audio for transcription" rather than "performance recordings only," tutorials with clean piano audio might be valid candidates. Worth deciding the intent explicitly before fixing the scoring logic, since the fix depends on whether tutorials are always-bad or only bad when they include non-piano content (talking, metronome clicks, etc.).

**3. Out-of-scope deletions of versioning/CI infrastructure**
`.github/workflows/release.yml`, `scripts/check_version_sync.sh`, `scripts/sync_pubspec_version.sh`, `frontend/lib/widgets/version_footer.dart`, `e2e/tests/pipeline-e2e.spec.ts` — all deleted.

These remove the automated release workflow, version sync guards, a user-visible version footer, and the only full-pipeline e2e test. None are related to cover search. **This may be a rebase artifact** — if these were already removed on `main`, rebasing will resolve it. If not, this should be split into a separate PR.

---

### Important

**4. `_preferCleanSource` not reset on mode switch**
`frontend/lib/screens/upload_screen.dart:245-249`

When the user switches away from YouTube mode, `_pickedFile` and `_error` are reset but `_preferCleanSource` is not. If the user toggles it ON, switches to Audio, then back to YouTube, the toggle is still ON — stale state from a prior mode session.

Fix: add `_preferCleanSource = false` inside `onSelectionChanged`.

**5. `from_title_lookup` builder silently drops `prefer_clean_source`**
`backend/services/ingest.py:360-366`

The static builder hardcodes `InputMetadata(...)` without the `prefer_clean_source` parameter. Production `POST /v1/jobs` doesn't use this builder so it's not a live bug, but any developer using it for tests or tooling will get the flag silently forced to `False`.

**6. `prefer_clean_source` unconditionally serialized for all job types**
`frontend/lib/api/client.dart:75-82`

The field is always included in the POST body, even for audio/MIDI uploads where it's meaningless. Not a runtime bug, but could be cleaned up: `if (preferCleanSource) 'prefer_clean_source': preferCleanSource,`.

---

### Test gaps

- No test for positive-piano-keyword + bad-keyword co-occurrence (would expose #2 — if you decide tutorials should be penalized)
- No test for `_maybe_swap_for_cover_sync` returning a non-YouTube-parseable URL (would expose #1)
- No test for toggle state reset on mode switch (would expose #4)

---

### Confirmed clean

- Config gating logic (kill switch + per-job opt-in) ✓
- Contract field `prefer_clean_source: bool = False` backward compatible ✓
- `POST /v1/jobs` correctly defaults to `False` for existing clients ✓
- Frontend YouTube URL regex handles all standard formats ✓
- Toggle visibility correctly gated to YouTube mode only ✓

jack0lantern (commented):
## Re-review after rebase

The rebase onto main resolved the out-of-scope CI/versioning deletions — the diff is now clean, 11 files, all cover-search-related. Updated status of all previously flagged issues:

---

### Resolved

**~~Out-of-scope deletions~~** — Fixed by rebase. The versioning infra, e2e tests, and release workflow are now in the base branch.

**~~Scoring: "solo piano" + "tutorial" co-occurrence~~** — **False positive in the original review.** The arithmetic was wrong: +30 (piano keyword) + 20 (title match) − 20 (tutorial penalty) = **30**, which does NOT clear the 60 threshold. The only way a tutorial clears is with an allowlist channel bonus (+50 → total 80), but those are trusted channels (Rousseau, etc.), so that's arguably correct behavior.

On the broader design question: should tutorials be penalized at all? A video like "Bohemian Rhapsody Solo Piano Tutorial" still contains a human playing piano — the audio track is exactly what Basic Pitch needs. The −20 penalty assumes tutorials are bad transcription sources, but the actual audio quality may be equivalent to a "piano cover" video. Worth considering whether the penalty should only apply to tutorials that are likely to have non-piano audio (talking, metronome clicks, split-screen with notation overlays that indicate synthetic playback, etc.) — or whether the current blanket −20 is the right trade-off for simplicity.

---

### Still present

**1. (Critical) Bare-ID cover URL can crash the ingest job**
`backend/services/cover_search.py:284-288` / `backend/services/ingest.py:192`

Still unguarded. `_yt_dlp_search` promotes `webpage_url` → `url` when `url` is absent, but does NOT handle yt-dlp returning a bare video ID string (e.g. `"dQw4w9WgXcQ"`) in the `"url"` field — which `extract_flat=True` is known to do. The bare ID flows through `CoverSearchResult.url` → `_maybe_swap_for_cover_sync` return → `_download_youtube_sync` → `extract_youtube_id()` which fails the hostname check → `ValueError` → job crash.

The silent-failure catch in `_maybe_swap_for_cover_sync` only wraps `find_piano_cover()`, not the downstream download. No test covers this path.

**2. (Important) `_preferCleanSource` not reset on mode switch**
`frontend/lib/screens/upload_screen.dart:245-249`

`onSelectionChanged` resets `_pickedFile` and `_error` but not `_preferCleanSource`. Toggle ON → switch to Audio → switch back to YouTube → toggle is still ON silently.

**3. (Important) `from_title_lookup` builder drops `prefer_clean_source`**
`backend/services/ingest.py:360`

The static builder doesn't accept the flag. The live API route constructs `InputBundle` directly so production is fine, but the builder is a silent footgun for tests/scripts/future callers.

**4. (Minor) `prefer_clean_source` unconditionally serialized for all job types**
`frontend/lib/api/client.dart:81`

Always included in POST body even for audio/MIDI where it's meaningless. Not a runtime bug — the backend ignores it — but semantically noisy.

---

### Missing test coverage

- No test for `_maybe_swap_for_cover_sync` receiving a non-YouTube-parseable URL (directly corresponds to issue #1)
- No test for toggle state reset on mode switch (corresponds to issue #2)

robin-raq:
## Review response

All four flagged issues addressed — plus two additional silent-failure modes you mentioned separately (MIDI byte validation, nonexistent URI acceptance). The branch now contains piano-tier + chiptune variants + Kesh Piano Music + the fixes below. **468/468 backend + 55/55 frontend** passing on `54abe8d`.

### 1. (Critical) Bare-ID cover URL can crash the ingest job — FIXED

Confirmed: the crash path was exactly as described. `extract_flat=True` sometimes puts the 11-char video ID directly in the `url` field, which flowed unchecked through `CoverSearchResult.url` → `_maybe_swap_for_cover_sync` → `_download_youtube_sync` → `extract_youtube_id`, where `urlparse` returned an empty hostname and `_download_youtube_sync` raised `ValueError` — violating the silent-failure contract.

**Two-layer fix:**

- **At the source** (`backend/services/cover_search.py`): new `_normalize_entry_url()` helper handles all three yt-dlp entry shapes we've observed. Preference order: `webpage_url` → `url` (only if it looks like a real URL) → reconstructed `https://www.youtube.com/watch?v={id}` from `id` → empty string (entry dropped). `_yt_dlp_search` now calls it on every entry and drops anything unresolvable.

- **Defense-in-depth** (`backend/services/ingest.py`): `_maybe_swap_for_cover_sync` validates `match.url` with `is_youtube_url()` before propagating. If a pathological entry ever slips through the normalizer, we fall back to the original URL and log a warning — the job continues instead of crashing.

**New tests** (8 + 1):
- `TestNormalizeEntryUrl` covers all three variants + the integration with `_yt_dlp_search` dropping unresolvable entries
- `test_falls_back_when_cover_result_has_non_youtube_parseable_url` exercises the defense-in-depth path directly with a forged `CoverSearchResult` containing a bare ID

The 4 existing ingest tests that used placeholder strings like `ROUSSEAU_COVER` and `originalVID` were updated to use real 11-char YouTube IDs — the new defense-in-depth layer correctly rejects the placeholders as non-parseable, which is exactly what it should do in production.

### 2. (Important) `_preferCleanSource` not reset on mode switch — FIXED

`frontend/lib/screens/upload_screen.dart:245` — `onSelectionChanged` now resets `_preferCleanSource = false` on every mode change. Added a Flutter widget test (`toggle resets to OFF after switching modes away and back`) that flips the toggle on in YouTube mode, switches to Audio, switches back, and asserts the submitted body contains `prefer_clean_source: false`.

### 3. (Important) `from_title_lookup` builder drops `prefer_clean_source` — FIXED

`backend/services/ingest.py:373` — the static builder now accepts `prefer_clean_source` as a **keyword-only** parameter (keyword-only specifically to prevent positional shadowing of `artist`, which would otherwise be a footgun for a 2-arg signature). Added `TestFromTitleLookupBuilder` with 3 tests: default is `False`, explicit `True` lands on metadata, positional argument raises `TypeError`.

### 4. (Minor) `prefer_clean_source` unconditionally serialized — FIXED

`frontend/lib/api/client.dart` — `preferCleanSource` parameter is now `bool?` (nullable) with default `null`. The body dict uses a conditional include: `if (preferCleanSource != null) 'prefer_clean_source': preferCleanSource`. The upload screen's YouTube branch still passes `preferCleanSource: _preferCleanSource` (a non-null bool), so its request bodies look identical to before. Audio / MIDI / title-mode branches don't pass the parameter at all, so the field is omitted from those request bodies.

Added a Flutter test (`title-mode submission does not include prefer_clean_source`) that submits a title-lookup job and asserts `containsKey('prefer_clean_source') == false`.

---

### Design question: should tutorials be penalized at all?

Thoughtful point — and I think the current design actually already does what you're suggesting, just through a different mechanism. Here's my reasoning for keeping the blanket `-20` over doing per-tutorial audio-content detection:

**What the penalty is really gating.** The `-20` is specifically targeting the worst-case tutorial: voiceover explaining technique, 50% speed demonstrations with finger overlays, chopped segments, metronome clicks, split-screen with a sheet-music sidebar where the audio pauses whenever the narrator is talking. For anonymous tutorial uploaders these patterns are very common, and Basic Pitch on that kind of audio produces dense garbage. The penalty is a cheap heuristic that says "absent positive signal, assume the worst."

**Why it still works for clean tutorials.** The exemption mechanism (`TRUSTED_TUTORIAL_CHANNELS` — see commit `5c069f2`) is the escape hatch. Kesh is the canonical example: 400+ videos all labeled "Piano Tutorial + MIDI" but every single one is pristine Synthesia-rendered piano with zero voiceover. Instead of trying to detect that from metadata we don't have (yt-dlp doesn't give us "has voiceover" or "is split-screen"), we hand-maintain a small allowlist. A Kesh match still scores well above threshold because:
- +50 for the allowlist match (`PIANO_MODERATE_CHANNELS`)
- +20 for title substring
- +10 for artist substring  
- 0 for the tutorial penalty (exempted)
- = 80, comfortably above the 60 threshold

**Why per-video audio inspection is the wrong trade-off right now.** Detecting "talking vs. rendered piano" from a YouTube metadata probe is basically impossible — we'd need to actually download and stem-separate the audio, at which point we're doing the entire transcription pipeline just to decide whether to use this source. That would defeat the whole purpose of cover_search as a fast-path router. The allowlist + exemption-list combination gives us the same end result (Kesh-like channels pass, no-name tutorial uploads get dinged) at zero audio cost.

**Where I'd revisit this.** If we start seeing a class of non-allowlisted, non-Kesh channels that clearly produce clean tutorial audio and keep getting unfairly dinged, the right move is to add them to `TRUSTED_TUTORIAL_CHANNELS` (a 1-line change) rather than loosen the heuristic for everyone. The allowlist has the virtue of being auditable — we can explain exactly why a channel's score went up.

Happy to revisit this if you see a specific case where the `-20` is causing a real miss.

---

### Additional fixes in this push (separate commit `54abe8d`)

Noticed the two silent-failure modes you raised during review:

- **`/v1/uploads/midi` accepted arbitrary bytes.** The route only checked the filename extension; a client could upload `song.mid` containing plain text or any other bytes and get a 200 OK. Fixed by checking for the Standard MIDI File magic header (`MThd`) at the route layer — garbage never touches blob storage. 4 new tests in `test_uploads.py`.

- **Jobs with nonexistent URIs silently succeeded.** `POST /v1/jobs` accepted any `RemoteAudioFile` / `RemoteMidiFile` without verifying the blob URI existed, so forged payloads ran through stub stages and "succeeded" with empty output. Added `LocalBlobStore.exists(uri)` (safely handles non-`file://` schemes, path-escape attempts, and missing files uniformly) and wired it into `create_job` as a 400 gate. 2 new tests in `test_jobs.py`.

---

### Commit layout

- `e6b61ed` … `19fc5d2` — original piano-tier fast path (unchanged from before)
- `e3cd5e6` — tiered piano allowlist + chiptune variant + `find_clean_source`
- `974311c` — ingest + frontend wiring for the chiptune variant
- `5c069f2` — Kesh Piano Music via `TRUSTED_TUTORIAL_CHANNELS` exemption
- `1f6a3bf` — **PR #47 review fixes** (issues 1-4 above)
- `54abe8d` — **MIDI magic header validation + URI existence check**

468 backend tests / 55 frontend tests / ruff clean.

shoyu-ramen (dismissed):

jack0lantern (dismissed):

jack0lantern (approved):

PR #60, opened 2026-04-13
Fix some bugs

swifttarrow:
## Description
Seeing some errors when I use certain inputs.

`celery.exceptions.MusicXMLExportException: Cannot convert inexpressible durations to MusicXML.`|
`celery.exceptions.MusicXMLExportException: Cannot convert "2048th" duration to MusicXML (too short).`

### Engrave / MusicXML

Adds duration coercion so floats and makeNotation-generated tuplets (e.g. 2048th “normal” types) do not trip music21 → MusicXML export (typeToMusicXMLType / MusicXMLExportException).
Runs splitAtDurations(recurse=True) (twice, with coercion between passes) so makeNotation=False export does not die on complex / inexpressible durations.
Writes MusicXML with makeNotation=False after explicit per-part notation, avoiding a second makeNotation pass inside the exporter that could reintroduce unwritable tuplets.

### Dev / storage

Dockerfile.dev: installs shared in editable mode (pip install -e ./shared) so bind mounts pick up host changes.
docker-compose.yml: bind-mounts ./shared into the relevant services so orchestrator code and shared stay in sync in dev.
backend/storage/local.py: LocalBlobStore subclass that always exposes exists() for job integrity checks—delegates to the shared wheel when present, otherwise implements a safe fallback (handles older installed ohsheet-shared without exists).

### Tests

New tests/test_engrave_quality.py cases for 2048th tuplet brackets, bare 2048th clamping, inexpressible float QL snapping, and the makeNotation=False + splitAtDurations regression.

### DIFF
--- a/Dockerfile.dev
+++ b/Dockerfile.dev
@@ -7,7 +7,8 @@ RUN apt-get update && apt-get install -y --no-install-recommends ffmpeg gcc g++ 
     && rm -rf /var/lib/apt/lists/*
 
 COPY shared/ shared/
-RUN pip install --no-cache-dir ./shared
+# Editable so bind-mounting ./shared in docker-compose tracks host changes.
+RUN pip install --no-cache-dir -e ./shared
 
 COPY pyproject.toml .
 COPY backend/ backend/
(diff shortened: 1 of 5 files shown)


jack0lantern (approved):

jack0lantern:
Approved but unable to merge. Failing some merge requirement. Maybe it needs an update from qa?

PR #13, opened 2026-04-08
feat(transcribe): waveform-derived tempo_map for audio jobs

shoyu-ramen:
## Summary

- Adds `backend/services/audio_timing.py` with `build_tempo_map_from_beat_times` and `tempo_map_from_audio_path`. Runs `librosa.beat.beat_track` on the waveform and converts detected beat instants into a piecewise `TempoMapEntry` list — one anchor per beat, `bpm = 60/Δt`, clamped to `[30, 300]`.
- Wires the override into `TranscribeService._run_basic_pitch_sync` → `_pretty_midi_to_transcription_result` so `HarmonicAnalysis.tempo_map` reflects the real pulse of the recording instead of a single global BPM from `pretty_midi.estimate_tempo`. Appends a `tempo_map from audio beat tracking (librosa)` line to `quality.warnings` when the override is used.
- Falls back silently to the prior single-anchor map when `librosa` is missing, beat tracking fails, the audio is too short (< 0.5 s), or no beats are detected. MIDI-upload jobs are unaffected — `pretty_midi` tempo changes already build a multi-point map there.

## Why

`ArrangeService` runs `sec_to_beat` in beat space, so small tempo errors compound into noticeable quantization drift against what the listener hears. This closes that gap for the audio path.

## Stacking

Based on `feat/basic-pitch-transcription` (PR #12) since this depends on the Basic Pitch transcription path being live. Re-target to `main` once #12 merges.

## Test plan

- [x] `pytest tests/test_audio_timing.py` — 6 new unit tests covering uniform 120 BPM continuity/monotonicity, variable tempo segment detection, empty/single-beat fallback, bpm clamping at both extremes, fallback bpm clamping, unsorted input sorting. No `librosa` required in CI.
- [x] `pytest` — full suite, 42 passed.
- [x] `ruff check backend tests` — clean.
- [x] `mypy backend` — clean.
- [ ] Manual smoke test on a real audio job once merged into a branch with Basic Pitch live: verify `quality.warnings` contains the beat-tracking line and arranged score beat positions follow the waveform pulse.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #62, opened 2026-04-14
feat: cover search improvements + external transcription service integration

robin-raq:
## Summary
- **Cover search hardening**: Expand allowlist with 16 new piano channels, block problematic channels whose arrangements crash engrave, filter out short videos (<60s), split artist from YouTube titles for better matching
- **External transcription delegation**: YouTube URL and title search jobs delegate to an external transcription service that produces cleaner scores. Audio/MIDI uploads still use the internal pipeline
- **Frontend cleanup**: Remove cover search toggle — always prefer clean source for YouTube. Embed external score viewer as iframe on result page
- **Score metadata**: Pass title/artist so scores render with correct song information

## Test plan
- [x] 584 backend tests pass (0 failures)
- [x] Cover search 100% hit rate across 50+ songs (multiple genres)
- [x] End-to-end: YouTube URL → cover search → transcription → result page with embedded viewer
- [ ] Team review
- [ ] Manual test on deployed qa environment


jack0lantern:
## Review — cover search + TuneChat integration

Substantial PR (+871/−1356). High-signal issues below, by severity.

### 🔴 Blocking / High

1. **`source_url` wiring mismatch (frontend ↔ backend)** — `backend/api/routes/jobs.py` adds `source_url` to `JobSummary` top-level, but `frontend/lib/screens/result_screen.dart:848` reads it from the `EngravedOutput` result map:
   ```dart
   _sourceUrl = resultMap['source_url'] as String? ?? ...
   ```
   That key lives on `JobSummary`, not inside `result`. The Dart `JobSummary` model (`frontend/lib/api/models.dart`) doesn't appear to have a matching `sourceUrl` field in this diff. The fallback to `job.title.startsWith('http')` also fails because `title` has been replaced with the resolved song name by the runner. Net: the "View on YouTube" link will never render. Either add `sourceUrl` to the Dart `JobSummary` model (and read `widget.job.sourceUrl`), or move `source_url` into `EngravedOutput`.

2. **`PipelineConfig.skip_humanizer` default flipped `False` → `True` globally** (`shared/shared/contracts.py`). This silently skips humanization on *every* pipeline variant (audio upload, midi upload, full), not just the TuneChat path. The two test changes in `tests/test_celery_dispatch.py` forcing `skip_humanizer=False` confirm the blast radius — any caller constructing `PipelineConfig()` without that arg now gets a degraded pipeline. If the intent is "skip humanizer when TuneChat takes over", gate it at the runner or at `get_execution_plan`, not at the contract default.

3. **Engrave defensive code removed while the internal pipeline still runs for audio/MIDI uploads** — `backend/services/engrave.py` loses 148 lines including `_coerce_durations_for_musicxml_export`, `_approximate_ql_for_musicxml_export`, `splitAtDurations(recurse=True)`, and the `makeNotation=False` export flag. The existing comments explicitly warn these fix `inexpressible`/`complex`/`2048th`-tuplet MusicXML export crashes. YouTube jobs now go to TuneChat, but audio/MIDI uploads still hit music21 engrave. What regression testing confirmed those crashes no longer reproduce? Accompanied by an **878-line deletion of `tests/test_engrave_quality.py`** with no replacement — that's the exact safety net for this removal.

4. **TuneChat path is sequential, not parallel** — the PR description says "run in parallel, upgrade when arrives", but `backend/jobs/runner.py` does a synchronous `await transcribe_via_tunechat(...)` inside the ingest step and *early-returns*, replacing the Oh Sheet pipeline entirely for `title_lookup`. The trailing comment `# Await the parallel TuneChat request (if it was fired).` is misleading — nothing is fired in parallel. Either reconcile description/comment with the actual "TuneChat-only early return" design, or implement the parallel behavior described.

### 🟡 Medium

5. **`score_pipeline` default drift** — `backend/config.py` Settings default flips to `"condense_transform"`, but `PipelineConfig.score_pipeline` default in `shared/shared/contracts.py` is still `"arrange"`. Depending on which path constructs `PipelineConfig`, downstream execution could disagree with the documented default.

6. **`get_execution_plan` replaces `arrange` with just `[condense]`**, dropping `transform`. Comment says transform is a no-op stub, so the removal is fine — but then the literal name `ScorePipelineMode = "condense_transform"` is lying to readers. Rename to `condense_only` or keep `transform` in the plan for symmetry/future work.

7. **Ingest title/artist split is fragile** — `backend/services/ingest.py` splits only on `" - "` (ASCII hyphen-space). YouTube titles commonly use `" – "` (en-dash), `" — "` (em-dash), `" | "`, or `"Song · Artist"`. Many will fall into the "no separator, artist=None" branch and degrade cover-search scoring. Consider a small multi-separator split or a dedicated parser.

8. **Lowercasing at ingest boundary** — `find_clean_source(title.lower(), artist.lower() if artist else None, ...)` pre-lowercases even though `normalize_title` / scoring already normalize. Harmless but redundant and couples layers; drop the `.lower()` calls or remove the internal normalization, not both.

9. **Docker dev-parity reversal** — `Dockerfile.dev` drops `-e ./shared` and `docker-compose.yml` drops the `./shared` bind mounts. Commit `ca478bb fix(docker): mount ./shared into worker-refine for dev parity` just restored the opposite. This PR undoes that for all workers. Intentional coordination with the assembler/decomposer services, or a merge collision?

### 🟢 Nits

10. **`upload_screen.dart` orphan comment** — the `setState` body retained a fragment ending mid-sentence: `// means the user could flip it ON, switch` with no completion. Either finish the thought or drop the whole block.

11. **Broken docstring in `tests/test_cover_search.py`** (chiptune-paused test): `# Chiptune variant is paused because TuneChat's pipeline` / `# The downstream transcription pipeline only handles piano...` — first sentence is cut off.

12. **`_NOISE_PATTERNS` regex `\s*-\s*official\s+\w+(\s+\w+)?`** is ambiguous about "Official Music Video" (3 words). Minor given the bracketed catch-all handles most cases.

13. **`tunechat_client.py` — `data` is reassigned** (`data = {}` for form fields, then `data = response.json()`). Readability — rename the form dict to `form_data`.

14. **`backend/storage/local.py`** is now a bare re-export; the prior docstring explained why the subclass existed (Docker bind-mount workaround). Given #9 the workaround is unneeded — consider importing from `shared` directly instead of keeping the shim.

### Recommendation

**Request changes**, primarily on issues 1–4. The product story (delegate YouTube to TuneChat, keep Oh Sheet for direct uploads) is clean, but the blast radius of the `skip_humanizer` default flip (#2) and the removed engrave safety net (#3) are worrying for the audio/midi-upload path that *doesn't* get TuneChat. The frontend `source_url` bug (#1) is a simple fix but ships as a broken user-visible feature.

robin-raq:
Thanks for the thorough review! All issues addressed in `f93b6f1` and `0bd24f5`:

**🔴 Blocking (1–4):**
- **#1** `source_url` wiring — added `sourceUrl` to Dart `JobSummary` model, now reads from job top-level. "View on YouTube" link renders correctly.
- **#2** `skip_humanizer` default reverted to `False`. TuneChat path already early-returns before humanize runs, so the global default was unnecessary.
- **#3** Engrave defensive code fully restored — `_coerce_durations_for_musicxml_export`, `_approximate_ql_for_musicxml_export`, `splitAtDurations(recurse=True)`, and `makeNotation=False` export flag. Also restored `test_engrave_quality.py` (878 lines).
- **#4** Removed stale "Await the parallel TuneChat request" comment. The TuneChat path is sequential (early-return after ingest) — comments now match behavior.

**🟡 Medium (5–9):**
- **#5/#6** `score_pipeline` renamed `condense_transform` → `condense_only` to reflect actual execution plan. Default aligned.
- **#7** Title split now uses `re.split(r"\s+[-–—|]\s+")` to handle en-dash, em-dash, and pipe separators.
- **#8** Redundant `.lower()` calls removed from ingest boundary — normalization happens inside cover_search scoring.
- **#9** Docker dev-parity restored — `Dockerfile.dev` and `docker-compose.yml` shared bind mounts back in place.

**🟢 Nits (10–14):**
- **#10** Orphan comment fragment in `upload_screen.dart` removed.
- **#11** Broken docstring in `test_cover_search.py` chiptune test fixed.
- **#12** `_NOISE_PATTERNS` regex — acknowledged, existing bracketed catch-all handles the "Official Music Video" 3-word case.
- **#13** Renamed ambiguous `data` → `form_data` in `tunechat_client.py`.
- **#14** Added explanatory docstring to `backend/storage/local.py` shim documenting why it exists.

jack0lantern:
## PR #62 Review — cover search + TuneChat integration

> **Heads-up:** this branch is out of date with `qa` and needs a merge before it can ship.

### Overview
Expands cover-search allowlist (advanced tier re-enabled, 16 new channels, blocklist, short-video filter, tutorial keyword re-signal), delegates `title_lookup` jobs to an external TuneChat transcription service, swaps result screen to an embedded TuneChat iframe for those jobs, and flips default `score_pipeline` from `arrange` → `condense_only`.

### Blocking / high-priority

- **Broken downloads on TuneChat path** (`backend/jobs/runner.py`, `frontend/lib/screens/result_screen.dart`): the TuneChat short-circuit returns `EngravedOutput` with `pdf_uri=""`, `musicxml_uri=""`, `humanized_midi_uri=""`. The result screen hides OSMD but still renders PDF/MIDI/MusicXML download buttons — clicking them will hit `/artifacts/.../pdf` and 404. Either hide the buttons when `_hasTuneChat`, or skip the download row entirely.
- **Breaking config change**: `ScorePipelineMode` literal renamed `condense_transform` → `condense_only` AND the default flipped from `arrange` → `condense_only`. Any existing `OHSHEET_SCORE_PIPELINE=condense_transform` in deployed env/`.env` will now fail pydantic validation at startup. Worth a migration note and probably keeping `condense_transform` as an alias for one release.
- **`backend/storage/local.py` shim regression**: the old subclass existed specifically because Docker Compose bind-mounts only `./backend`, so route code can run against an older `ohsheet-shared` wheel lacking `exists()`. Replacing with `from shared.storage.local import *` removes that fallback. Please verify the deployed wheel definitely ships `exists()`, or restore the shim.

### Medium

- **TuneChat short-circuit lives inside the `ingest` branch of the stage loop** (`runner.py`). It synthesizes `stage_started`/`stage_completed` events with hardcoded progress values for stages that never actually ran (`transcribe`, `engrave`). Two concerns: (1) frontend subscribers may expect certain artifact events they won't get; (2) the fake progress pings are fragile and will drift from any future event-shape change. Consider factoring this into a dedicated variant/executor rather than a conditional mid-loop with early return.
- **Artist splitter is brittle** (`backend/services/ingest.py`): `re.split(r"\s+[-–—|]\s+", raw_title, maxsplit=1)` treats the first segment as artist. Titles like `"Song Name - Live at Wembley - Artist"` or `"Artist feat. X - Song"` will mis-assign. Low impact (falls back to keyword search), but worth a comment acknowledging the heuristic. Also, `import re as _re` belongs at module scope.
- **Noise pattern broadened too far** (`cover_search.py`): `r"\s*-\s*official\s+\w+(\s+\w+)?"` will eat trailing text in titles like `"Song - official release announcement"`. Previous version constrained to `official (music )?video`. Consider anchoring to the known variants.
- **Test weakened** (`tests/test_cover_search.py`): `test_cover_search_min_score_defaults_to_60` → `is_sensible (20..80)`. Original intent was to catch drift between `config.py` and `cover_search.py`; the new range assertion won't. Restore the strict check or add a separate drift test.

### Minor

- **UX regression not called out in PR description**: piano-roll `listenSection` is deleted from the result screen entirely. Intentional?
- **`iframe.src` contains user-provided title/artist** in query params (`result_screen.dart`). `Uri.replace(queryParameters: ...)` percent-encodes, so no injection — just confirming it's URL-encoded correctly in TuneChat's router.
- **`tunechat_timeout_sec: int = 300`** = 5 min blocking await inside the job runner. For a Celery-dispatched pipeline this is fine, but document that jobs now pin a worker for up to 5 minutes on the TuneChat round-trip.
- **`noqa: BLE001` triple** in runner + client is fine given the silent-failure contract, but add a single catch-all comment at top of the block rather than per line.

### Looks good
- Cover-search scoring changes have matching test updates (allowlist bonuses, `piano tutorial` positive keyword, chiptune paused path, blocklist/duration filters).
- `tunechat_client.py` has a clean silent-failure contract and correct error classification.
- Frontend refactor into `StatefulWidget` with deferred viewer + `RepaintBoundary` is a reasonable perf win.
- `source_url` propagation end-to-end (contracts → API → Dart model → view-on-YouTube link) is consistent.

### Security
No new auth/injection surface. TuneChat API key is read from settings (not logged). Iframe `src` is constructed from a trusted env-configured base + encoded query params.

### Recommendation
Request changes — fix download buttons on TuneChat path, resolve the `condense_transform`/shared-wheel breaking changes, and merge `qa` into the branch before shipping.

robin-raq:
## Second-pass review fixes — pushed in `6171bf0`

All blocking, medium, and nit items from the [second-pass review](https://github.com/Oh-Sheet-Team/oh-sheet/pull/62#issuecomment-4257107208) addressed.

### 🔴 Blocking

1. **Broken downloads on TuneChat path** → `frontend/lib/screens/result_screen.dart`. Wrapped the downloads row in `if (!_hasTuneChat)` so the buttons (and their spacer) are skipped entirely when the TuneChat iframe is active. TuneChat's iframe ships its own download controls.

2. **`condense_transform` breaking rename** → `backend/config.py`. Added a `mode="before"` field validator on `Settings.score_pipeline` that maps the deprecated string to `condense_only` and emits a `DeprecationWarning`. Deployed `.env` files with the old value keep booting; operators get a clear upgrade signal. The `PipelineConfig` literal stays clean with only `condense_only`.

3. **Storage shim regression** → `backend/storage/local.py`. Restored the defensive subclass of `shared.storage.local.LocalBlobStore` with a fallback `exists()` impl, in case a future deployment bind-mounts `./backend` without `./shared`. Dev compose bind-mounts both today; this is belt-and-suspenders against drift.

4. **Prod safety pin** → `docker-compose.prod.yml`. Added `OHSHEET_SCORE_PIPELINE: arrange` to the orchestrator's env so the default flip can't silently change prod behavior on deploy. Only the orchestrator reads this (PipelineConfig is built at job-creation time in `api/routes/jobs.py`; workers just execute stages). Plan is to remove the env line in a follow-up PR after validating `condense_only` in prod on a few test jobs.

### 🟡 Medium

5. **Noise regex over-broad** → `backend/services/cover_search.py`. Replaced `r"\s*-\s*official\s+\w+(\s+\w+)?"` with an enumerated list:
   ```python
   r"\s*-\s*official\s+(music\s+video|lyric\s+video|video|audio|visualizer|mv)\b"
   ```
   Added `test_dash_official_pattern_does_not_overmatch` covering "Big Hit - Official release announcement" (must survive the normalization pass).

6. **Ingest module-scope `import re` + splitter docs** → `backend/services/ingest.py`. Moved `import re` to module scope (was `import re as _re` inside the function body). Added a comment acknowledging the heuristic's known limits: first-dash-wins splits "Song Name - Live at Wembley - Artist" as (Song Name, Live at Wembley - Artist), and there's no disambiguation between "Artist - Song" vs "Song - Artist" ordering.

7. **Restored strict `cover_search_min_score` drift test** → `tests/test_cover_search.py`. The prior test only asserted `20 <= settings.cover_search_min_score <= 80`, which wouldn't catch a drift between `Settings.cover_search_min_score` and `find_clean_source`'s signature default. Added `test_cover_search_min_score_matches_find_clean_source_default` that uses `inspect.signature` to compare the two defaults directly.

### 🟢 Nits

8. **Silent-failure comments at catch-all sites** → `backend/services/tunechat_client.py` + `backend/jobs/runner.py`. Added a leading comment above each `except Exception` block explaining the silent-failure contract, instead of just the per-line `# noqa: BLE001`. (Ruff requires the per-line directive, so the `noqa` stays — the comment above it documents intent.)

9. **Documented `tunechat_timeout_sec` worker pinning** → `backend/config.py`. Added a note that the default 300s timeout pins a Celery worker for up to 5 minutes per TuneChat round-trip, so operators should size worker concurrency accordingly when TuneChat is enabled.

### Deferred (not in this commit)

- **TuneChat short-circuit executor refactor** (medium #4 from review). Pulling the TuneChat short-circuit out of the ingest branch into a dedicated variant/executor is a real architectural change — should be its own PR, not bundled with review-fix plumbing. Tracked in ROADMAP.
- **Piano-roll `listenSection` deletion** (minor from review). Intentional — leaving it out while TuneChat owns the listening experience for `title_lookup` jobs.

### Post-rebase

Pulled `qa` → `feat/cover-search-tunechat` merge (`37739b1`) before pushing. Rebase applied cleanly, no conflicts. Ruff clean, mypy clean (on tracked files), 132 tests passing on the main suite.

jack0lantern:
## Re-review — all prior feedback addressed ✅

Walked through commit `6171bf0` and the `qa` merge. Every item from the earlier review is resolved cleanly:

### Blocking — all fixed
- **Download buttons hidden on TuneChat path** (`result_screen.dart:108`): `if (!_hasTuneChat)` gates the row with a clear inline rationale. No more 404s on click.
- **`condense_transform` deprecation alias** (`config.py:540`): `mode="before"` field validator coerces the old value and emits a `DeprecationWarning` instead of failing pydantic validation. Good migration path.
- **`backend/storage/local.py` shim restored**: defensive subclass with `exists()` fallback is back, with an updated docstring explaining the historical reason and why it's still kept despite the now-shared bind mount.
- **Prod pin via `OHSHEET_SCORE_PIPELINE=arrange`** (`docker-compose.prod.yml:32`): nice belt-and-braces — prevents the default flip from silently changing prod behavior. Comment correctly notes only the orchestrator reads it and that the pin should be removed after `condense_only` is validated in prod.

### Medium — all fixed
- **Noise regex narrowed** (`cover_search.py:247`): enumerated `(music\s+video|lyric\s+video|video|audio|visualizer|mv)\b` instead of the open-ended `\w+(\s+\w+)?`. Plus a regression test (`test_dash_official_pattern_does_not_overmatch`) locking in that "Official release announcement" survives.
- **`import re` hoisted to module scope** (`ingest.py:17`) and the artist-splitter heuristic now has an explicit "Heuristic limits (acknowledged)" docstring listing the known mis-split cases.
- **Drift test added** (`test_cover_search.py:531`): `inspect.signature(find_clean_source).parameters["min_score"].default` is asserted equal to `Settings.model_fields["cover_search_min_score"].default`. Restores the catch-on-drift property the previous strict-equality test had.

### Nits — fixed
- Silent-failure rationale comments at the catch-all `Exception` sites in `tunechat_client.py` and `runner.py` tunechat-only path.
- `tunechat_timeout_sec` now documents that it pins a Celery worker for up to 5 minutes — concurrency-sizing guidance is right next to the setting.

### Branch state
`qa` merged in (`37739b1`). `mergeStateStatus: BLOCKED` remains — likely waiting on required reviews/checks rather than conflicts (`mergeable: MERGEABLE`).

### Recommendation
**Approve.** Ship after the remaining required-check/review gates clear. Suggested follow-up after merge: file a tracking issue to remove the `condense_transform` alias and the `OHSHEET_SCORE_PIPELINE=arrange` prod pin in the next release once `condense_only` is validated.

jack0lantern (approved):

PR #64, opened 2026-04-14
feat: GAU-105 LLM refine stage + engrave notation fixes

jack0lantern:
## Summary

- **GAU-105 — LLM engrave refinement (new pipeline stage)**: Adds a `refine` stage between `humanize` and `engrave` that calls Claude via Anthropic tool-use to annotate score metadata (title, composer, key, tempo marking, sections, repeats) before engraving. Includes Celery worker, config, prompt/tool-schema builders, RefineService with retry/budget/fallback/cache, live golden-set eval harness, filename fallback when the song is unidentifiable, and ~20 tests.
- **Engrave notation fixes (bundled on same branch)**: Two-part braced grand staff (MusicXML `<part-group>` with brace), MusicXML corruption fix (4 interacting music21 quirks that MuseScore flagged as corrupt), per-part tie-chain scoping, 1/12 qL Fraction-based snap before export, regression tests.
- **Pre-PR review fixes**: 6 follow-up commits at the tip of this branch address 2 Critical + 8 Important issues from the reviewer pass (cache-key hint inclusion, `tool_choice` hardening, hint clamping, overlap-floor/snap-grid alignment, bar-split ε-before-bar drop, zero-duration handling, and helper-level unit tests).

## What's included

**New files (refine stage):**
- `backend/services/refine.py` — RefineService with retry/budget/fallback/cache
- `backend/services/refine_prompt.py` — pure prompt + tool-schema builders
- `backend/workers/refine.py` — Celery task
- `scripts/eval_refine.py` — live golden-set harness (real Anthropic API, gated by `make eval-refine`)
- `eval/fixtures/refine_golden/clair_de_lune/` — seed fixture
- `refine-baseline.json` — eval baseline (n=1, more to come)
- `.env.example` — refine-stage env config
- 8 new test modules under `tests/test_refine_*.py`

**Modified:**
- `shared/shared/contracts.py` + `backend/contracts.py` — schema v3.1.0 (refine fields, `RemoteAudioFile/MidiFile.source_filename`, `Repeat`, `ScoreSection`, tighter `staff_split_hint` bounds)
- `backend/services/engrave.py` — two-part braced grand staff; MusicXML corruption fixes; 1/12 qL snap; tie-chain per-part
- `backend/jobs/runner.py` — refine insertion in the pipeline plan; prefers refined title/composer
- `backend/api/routes/{jobs,uploads}.py` — refine_enabled flag, filename fallback
- `docker-compose.yml` / `docker-compose.prod.yml` — `worker-refine` service (amd64 platform, shared/ volume mount)
- `tests/test_engrave_quality.py` — two-part grand staff regressions + 12 helper unit tests

## Scope gaps (noted for follow-up, not blockers)

- **Engrave currently consumes only `title` and `composer`** from refined metadata. The other 5 refined fields (`tempo_marking`, `staff_split_hint`, `custom_label`, `repeats`, `arranger`) land on `ScoreMetadata` but don't yet reach the rendered PDF. File follow-up as Phase 2.
- **Eval baseline is n=1.** Design called for 10-15 fixtures; \`clair_de_lune\` is the seed. Expand in a follow-up.
- **Quintuplet/septuplet rhythms are silently snapped** by the 1/12 qL grid. Latent — arrange's current grid is LCM(3,4)=12 so unreachable today, but worth a ticket if quintuplet support is ever planned.
- **Minor items** (undocumented `PROMPT_VERSION=\"v2\"` bump, `_cache_uri` silently disabled for S3 blob stores, dead `last_exc` branch) — tracked in review findings.

## Review findings — addressed in this branch

See commits \`1415482\` through \`b2ff128\` for the pre-PR review-fix pass. Summary:

- **Critical**: cache key now hashes hints (was silently returning stale responses across different hints); `tool_choice={\"type\": \"any\"}` forces structured output (was accepting plain text).
- **Important**: `_clamp_hint` bounds prompt cost + injection surface; engrave overlap floor aligned with snap grid (was re-introducing the overlap it fixed); `_split_bar_crossing_notes` no longer silently drops notes starting ε-before a bar line or with zero/negative duration; \`test_refine_defaults\` now deterministic regardless of local `.env`; `test_llm_failure` renamed + split to explicitly cover non-transient and transient-exhaustion paths; missing `source_filename` round-trip test added; missing `./shared:/app/shared` volume mount on `worker-refine`.

## Test plan

- [ ] `pytest tests/` runs green — 648 passing, 83% coverage locally
- [ ] `pytest tests/test_refine_*.py tests/test_contracts_refine.py` — 28 tests
- [ ] `pytest tests/test_engrave_quality.py` — 109 tests (includes 12 new helper unit tests + 2 two-part grand staff regressions + bass-range LH clef regression)
- [ ] `pytest tests/test_uploads.py` — 9 tests (includes new source_filename round-trip)
- [ ] `docker compose up worker-refine` — boots on Apple Silicon (platform: linux/amd64), shared/ hot-reloads
- [ ] `make eval-refine` — live Anthropic API call against `clair_de_lune` golden fixture (requires `OHSHEET_ANTHROPIC_API_KEY`; costs real money — excluded from CI)
- [ ] End-to-end: upload an audio file via `POST /v1/uploads/audio`, create job with `enable_refine=true`, verify PDF shows refined title/composer


jack0lantern:
Waiting on QA fix verification before merging!

swifttarrow:
# Code Review: PR #64 — GAU-105 LLM refine stage + engrave notation fixes

**Scope:** 93 commits, 66 files, ~10,885 additions / 459 deletions. Bundles two features: a new LLM-driven `refine` stage between humanize and engrave, and a set of engrave/MusicXML notation fixes (two-part grand staff, OSMD/MuseScore corruption fixes). PR description notes 6 follow-up review-fix commits at the tip.

## Overview

- **New refine stage**: `RefineService` calls Claude (Anthropic tool-use + `web_search_20250305`) to annotate `ScoreMetadata` with title/composer/tempo_marking/sections/repeats. Integrated via a new Celery worker and pipeline-runner branch. Schema bumps to 3.1.0.
- **Engrave fixes**: Two-part `<part-group>` brace replaces `PartStaff`, per-part tie-chain scoping, Fraction-based 1/12 qL snap, and five post-export MusicXML sanitizers (voice remap, chord-tag collision fix, tie-chain voice alignment, part-name clear, UUID-id normalization).
- **Infra**: `worker-refine` service in both compose files; `anthropic>=0.40` added as a top-level dependency; `OHSHEET_ANTHROPIC_API_KEY` threaded through deployment.

## Strengths

- **Correct failure semantics.** Every error path in `refine.py` returns the input unchanged and appends a `QualitySignal.warnings` entry (for `HumanizedPerformance`). `PianoScore` inputs (no `warnings` field) pass through silently — that's a reasonable contract for a non-blocking enrichment stage.
- **Cache key is tight.** `_cache_key` now hashes the score, prompt version, model, and all three hints (title/artist/filename). The pre-PR fix for this avoids a correctness bug where filename-fallback responses would be stale.
- **`tool_choice={\"type\": \"any\"}`** forces a tool call so the model can't return a plain-text response. Good trade-off given you can't pin `submit_refinements` specifically without blocking the preceding `web_search`.
- **Hint clamping** (`_clamp_hint`) strips control chars and bounds length — shrinks both prompt cost and the prompt-injection surface.
- **Pure prompt module.** `refine_prompt.py` contains no I/O, only string/schema builders. This makes testing trivial (`tests/test_refine_prompt.py` has 165 lines).
- **Engrave sanitizers are well-isolated** and each one has a docstring explaining the music21/OSMD/MuseScore quirk it fixes — these will be valuable when a music21 upgrade breaks something.
- **Test coverage is substantial** — 28 refine tests, 109 engrave tests, and the PR-description claim of 648/83% tracks with what I see in the diff.

## Issues

### Correctness

- **`refine.py:217-219` — dead `last_exc` branch.** The retry loop always either `raise`s (line 191, when `attempt == len(backoffs) - 1` and exc is non-transient or last attempt) or `return`s. The trailing `if last_exc is not None: raise last_exc` is unreachable. PR description acknowledges it; fine to keep as defensive, but a comment saying so would help future readers.
- **`refine.py:405-413` — `_is_transient` uses string matching** on `\"timeout\"/\"overloaded\"/\"rate limit\"/\"connection\"`. Anthropic SDK exceptions have proper types (`APIConnectionError`, `APITimeoutError`, `APIStatusError`) — `isinstance` checks would be more robust. The `status_code >= 500` fallback covers the important case, but a rename/wording change in a future SDK release will silently break retry classification.
- **`_fix_voice_collision_chord_tags` (engrave.py:1001-1012)** relocates grace notes and mid-measure `<direction>` elements to the front of the measure. The in-code comment acknowledges this; for current arrange output it's a no-op, but it's a latent correctness bug if future phases emit mid-measure dynamics or pedal directions.
- **`refine_budget_sec=300` vs `refine_call_timeout_sec=120` × 3 attempts + 5s backoffs = up to ~365s per-call wall time.** The outer `asyncio.wait_for(timeout=budget)` does cap it, but the relationship is undocumented. If an operator raises `refine_call_timeout_sec` to 200 they'll silently never see the second retry. Worth a sentence in the config comment.

### Design / Maintainability

- **`PROMPT_VERSION = \"v2\"` bump is undocumented.** `refine_prompt.py:20` has no comment explaining what changed from v1 or why caches should invalidate. Future contributors will have no signal about when to bump it. Even a one-line changelog (\"v2: added filename-fallback defaults\") would help.
- **`_cache_uri` silently disables for non-LocalBlobStore** (refine.py:253-261) via `getattr(self.blob_store, \"root\", None)`. When S3/GCS is plumbed in, refine caching will be off with no log. A `log.debug` on first miss would surface the regression.
- **`_merge()` field whitelist is hand-maintained** (refine.py:299-354). If the tool schema gains a new field but `_merge` isn't updated, the LLM will happily return it and it'll be silently dropped. A symmetric helper (or a `log.debug(\"unknown refine field: %s\")`) would catch schema drift.
- **`worker-refine` in `docker-compose.yml`:96-100 builds its own image** (`build: Dockerfile.dev`) instead of using `\${BASE_IMAGE}` like every other dev service. Since `anthropic>=0.40` is a top-level dep (installed by `pip install .[...]` in `Dockerfile.dev:35`), the shared base image should contain it. This forces a cold local rebuild for no apparent reason and risks image drift from the other workers. Switch to `image: \${BASE_IMAGE:-...}` unless there's a specific reason otherwise.
- **Refine warnings never reach the API.** `RefineService._with_warning` appends to `HumanizedPerformance.quality.warnings`, but `JobSummary`/`EngravedOutput` don't expose `quality.warnings`. Users whose API key is absent or quota-exhausted silently get unrefined output with no signal. Either surface warnings in the job summary or leave a TODO.

### Scope

- Engrave consumes only `title` + `composer` from refined metadata. The other five fields (`tempo_marking`, `staff_split_hint`, `custom_label`, `repeats`, `arranger`) land on `ScoreMetadata` but aren't rendered. PR description flags this as Phase 2 — fine, but the feature ships with a visible capability-gap for anyone reading the tool schema.
- Eval baseline is n=1 (`clair_de_lune`). Design called for 10-15. Ship note / follow-up ticket is appropriate.

### Security

- `_clamp_hint` is a reasonable mitigation for prompt injection via user-provided title/filename. It's not a boundary — an adversarial filename can still alter Claude's behavior within 200 chars — but given Claude is only reading metadata, the blast radius is bounded to returning junk refined fields which `_merge`'s validators gate against. Acceptable.
- No obvious secrets handling regressions. `OHSHEET_ANTHROPIC_API_KEY` is correctly env-threaded and never logged.

### Performance

- `AsyncAnthropic` client is instantiated per task (`backend/workers/refine.py:48`) because `RefineService` is fresh each call. With the Celery prefork pool that's acceptable; if concurrency is ever increased or the pool moves to gevent, connection pooling will become a consideration.
- `_cache_key` serializes the entire `PianoScore` to JSON on every call. For a score with thousands of notes this is non-trivial (O(n) JSON encode + sha256). Fine for typical job volumes, worth noting if refine ever gets hit on large batches.

## Recommendation

Approve with minor follow-ups. The critical/important findings were already addressed in the pre-PR review pass (commits 1415482..b2ff128). Remaining items are documentation/quality-of-life polish and one small infra hygiene fix (`worker-refine` `build:` → `image:`). The feature is well-tested, fails safely, and the engrave sanitizers are appropriately defensive for the music21/OSMD/MuseScore pipeline's quirks.

**Suggested pre-merge tweaks:**
1. Add a one-line comment above `PROMPT_VERSION = \"v2\"` explaining what v2 changed.
2. Change `worker-refine` in `docker-compose.yml` to use `\${BASE_IMAGE}` like the other services.
3. Either surface `quality.warnings` in `JobSummary` or add a TODO comment noting refine failures are invisible to clients.

Everything else can land as follow-up tickets.

swifttarrow:
@jack0lantern Gonna approve. Some minor suggestions raised by Claude but nothing blocking!

swifttarrow (dismissed):

robin-raq (changes requested):
⚠️ The one thing I'd want surfaced before merge
OHSHEET_ANTHROPIC_API_KEY isn't plumbed through deploy.yml. This is exactly the class of bug we just fought with OHSHEET_TUNECHAT_* — PR #64 adds the var to docker-compose.prod.yml (environment: OHSHEET_ANTHROPIC_API_KEY: ${OHSHEET_ANTHROPIC_API_KEY:-}), but I don't see a corresponding addition to the Deploy to VM step's env: block or the /tmp/.env write in .github/workflows/deploy.yml.

Consequence on merge: worker-refine starts, tries to call Anthropic, gets "Unauthorized" or empty-key error, silently fails (the service correctly returns None on any exception), and every job's refine stage becomes a no-op. No alerts, no visible failure — same pattern as our TuneChat bug.

Verification quickly: git grep -i ANTHROPIC .github/workflows/deploy.yml on the PR branch. If it returns nothing, the env var won't reach the VM.

This is easy to fix in a follow-up (one GH secret set + 3 lines added to deploy.yml), but it needs to happen before anyone tests refinement end-to-end on qa/prod. Otherwise you'll hit the same silent-fail debugging spiral we just hit with TuneChat.

robin-raq:
Strengths worth calling out
Prompt-injection defense is thoughtful: _clamp_hint strips control chars, truncates to 200 chars, applied to title_hint/artist_hint/filename_hint. Cache-key includes these hints (addresses the Critical finding noted in the PR body).
tool_choice={"type": "any"} forces structured output — model can't fall back to plain text. Good fix for the other Critical finding.
Silent-failure contract maintained: LLM error → input passes through unchanged, no job crash. Consistent with TuneChat client's pattern.
MusicXML corruption fixes are documented with real "why": The _split_bar_crossing_notes comment explains music21's voice renumbering + OSMD sanitizer interaction + MuseScore 4's strict tie-chain validation in full detail. That's production-grade engineering documentation.
Test coverage: 6 dedicated refine test files (test_refine_{service,prompt,runner,config,integration,engrave_metadata}) — retry, budget, fallback, cache all covered.

Non-blocking observations (file as follow-ups if you want)
Prompt-version coupling: PROMPT_VERSION="v2" bumped without a documented reason (PR body even calls this out as "minor"). If SYSTEM_PROMPT text changes in a future PR without a version bump, cache serves stale responses. Cleaner: hash SYSTEM_PROMPT into _cache_key automatically.
Cache race: _cache_get uses exists() then get_json() — two concurrent jobs on the same song can both miss + both call Claude + both write. Low probability, low cost (~$0.01 extra API call), not blocking.
_cache_uri silently disabled for S3 stores (self-disclosed). If prod ever switches from LocalBlobStore to S3, refinement gets ~5× more expensive because cache becomes a no-op. Tracking ticket worth it.
5 refined fields land in metadata but don't reach the PDF (self-disclosed as Phase 2). Means half the LLM value is shelved until the engrave follow-up. Expected, but worth a Jira ticket so it doesn't languish.
Eval baseline n=1 (self-disclosed — design called for 10-15). Regressions could slip in with a golden set this small. Expand to ≥5 before shipping any subsequent refine-related changes.
Two features bundled (refine stage + engrave notation fixes). PR body calls this out. Reviewable as-is because they share a common fixture set, but I'd normally split. Fine here.

Questions :
Cost budget: each refine call is 1 web_search tool loop + 1 submit_refinements tool call. What's the expected $/job? config.refine_monthly_budget_usd would be nice — the service has retry limits but no budget ceiling I could see.
Fallback policy: if OHSHEET_ANTHROPIC_API_KEY is empty, should refine be a soft-disable or a hard error at worker startup? Currently soft-disables with per-job silent-failure, which is production-safe but operator-invisible.

robin-raq (approved):
Approving — re-reviewed after the pre-merge-fix commits.

**What resolved my earlier CHANGES_REQUESTED:**

1. `OHSHEET_ANTHROPIC_API_KEY` is now plumbed through `deploy.yml` in both the Deploy step's `env:` block and the `/tmp/.env` write (commit `0c8529f`). Same commit also aligned `worker-refine` in `docker-compose.yml` to `${BASE_IMAGE}` per swifttarrow's hygiene finding.

2. My underlying worry — that an empty key would cause silent per-job failures at the Anthropic call (the TuneChat bug class) — was based on a misread. Tracing the gating chain:
   - `backend/api/routes/jobs.py:162` sets `enable_refine=settings.refine_active` on every new job
   - `backend/config.py:510-518` — `refine_active` returns `anthropic_api_key is not None` when `refine_enabled` is unset
   - `shared/shared/contracts.py:405` — `get_execution_plan()` only inserts `refine` when `enable_refine=True`

   So with no key, the stage is dropped from the execution plan at job creation — never dispatched, never attempted. `worker-refine` boots but sits idle. That's cleaner than silent pass-through and exactly the right behavior for deferred enablement.

**Non-blocking follow-ups worth tickets (not re-raising here):**

- No `refine_monthly_budget_usd` ceiling — only spend control is Anthropic workspace-level caps if we create per-env workspaces
- Phase 2 metadata fields (`tempo_marking`, `staff_split_hint`, `custom_label`, `repeats`, `arranger`) accepted by schema but not rendered to PDF yet
- Eval baseline n=1 — expand `refine_golden/` before the next refine-related change
- `_cache_uri` silently no-ops for non-LocalBlobStore (raised in swifttarrow's review)
- `PROMPT_VERSION = "v2"` bump lacks a one-line changelog comment

**Engrave side:** swifttarrow already approved. The MusicXML sanitizers, two-part braced grand staff, per-part tie-chain scoping, and 1/12 qL snap fix real rendering bugs and ship value independent of refine.

**Operator note for anyone merging:** The `OHSHEET_ANTHROPIC_API_KEY` secret is not set in either the `qa` or `prod` GitHub environments. By design, this means refine auto-disables (see gating chain above) — the pipeline behaves exactly as before for the refine step. Flip the secret on via `gh secret set OHSHEET_ANTHROPIC_API_KEY --env {qa,prod} --repo Oh-Sheet-Team/oh-sheet` whenever we decide LLM refinement is worth the spend; the plumbing from this PR means it'll activate on the next deploy with zero code change.

robin-raq (approved):

PR #75, opened 2026-04-17
fix(ci): aggressive VM disk cleanup before docker pull

lramosve:
## Summary

PR #74 fixed the runner-side disk exhaustion (build+push now succeeds), but the **VM** ran out of disk during `docker compose pull` — the `until=24h` filter on `docker image prune` was too conservative and let recent deploy images accumulate.

**Change:** replace `docker image prune -af --filter 'until=24h'` + `docker builder prune -af` with `docker system prune -af` (both pre-pull and post-pull). This removes all unused images, build cache, stopped containers, and dangling volumes. Images backing running containers are safe from prune.

Failed run: https://github.com/Oh-Sheet-Team/oh-sheet/actions/runs/24570921209

## Test plan

- [ ] Merge and verify the next qa deploy succeeds end-to-end (both build and VM pull)

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -121,24 +121,18 @@ jobs:
           $SCP /tmp/.env "${VM_USER}@${VM_HOST}:~/oh-sheet/.env.tmp"
           $SSH "mv ~/oh-sheet/.env.tmp ~/oh-sheet/.env && mv ~/oh-sheet/docker-compose.prod.yml ~/oh-sheet/docker-compose.yml"
 
-          # Reclaim disk BEFORE pulling the new image. The app image ships
-          # heavy ML deps (torch/transformers via the [pop2piano] extra),
-          # so a single layer can exceed 1.6 GB. Without an aggressive
-          # pre-pull prune, stale tagged images from prior deploys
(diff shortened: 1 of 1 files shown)


robin-raq (approved):
Approved — this is the right call.

The `docker system prune -af` approach is simpler than both the original `until=24h` filter and my #73 attempt to tighten it to `until=1h`. Key correctness point: Docker protects images backing *running* containers from prune regardless of flags, so the `until=...` filter was always belt-and-suspenders — this removes the indirection.

Also worth noting the self-healing property: the current qa VM is stuck because nothing was old enough to satisfy the old `until=24h` filter. `system prune -af` has no such constraint, so the first deploy after this merges should free the stuck disk automatically — no manual VM prune required.

Minor nits (not blocking):
- Lost the `run 24407031842 / 24547994092 / 24570921209` incident IDs from the comment. Archaeology still available via git blame/log.
- Rollback semantics silently change: every rollback now requires `docker compose pull` from Artifact Registry rather than re-tagging a local image. Fine for qa; worth a second look if/when the same workflow runs against a prod VM with stricter rollback SLAs.
- `--volumes` intentionally NOT added — correct, preserves `blob-data`/`redis-data`/`caddy-data`.

Will close my #73 as superseded.

PR #78, opened 2026-04-17
feat(ingest): yt-dlp cookies via OHSHEET_YTDLP_COOKIES secret

robin-raq:
## Summary

YouTube has been flagging the qa/prod VM's IP as bot traffic and returning \"Sign in to confirm you're not a bot\" from yt-dlp, breaking the ingest stage for real users (observed in production today on the URL \`youtube.com/watch?v=tRrxcL-sSQQ\`).

This PR plumbs cookies from a logged-in browser session through the deploy pipeline so yt-dlp authenticates its requests and drops out of the bot-detection bucket.

## Architecture (mirrors OHSHEET_TUNECHAT_API_KEY plumbing)

```
GH secret OHSHEET_YTDLP_COOKIES   ← new, set via: gh secret set OHSHEET_YTDLP_COOKIES --env {qa,prod}
      │
      ▼ deploy.yml writes to /tmp + SCPs to VM
~/oh-sheet/youtube-cookies.txt (chmod 600)
      │
      ▼ docker-compose.prod.yml bind-mount (:ro)
/app/youtube-cookies.txt (inside orchestrator + worker-ingest)
      │
      ▼ OHSHEET_YTDLP_COOKIES_PATH env
settings.ytdlp_cookies_path
      │
      ▼ _ytdlp_utils.apply_ytdlp_cookies(ydl_opts)
ydl_opts[\"cookiefile\"]   ← only set when file exists AND size > 0
```

## Safety properties

- **Unset secret is a safe no-op.** The deploy writes an empty cookies file; the bind-mount still works; services detect size 0 and skip the cookiefile arg; yt-dlp runs anonymously. No breakage on VMs that haven't been provisioned yet.
- **Cookies file protected** with \`chmod 600\` on the VM and \`:ro\` mount into containers.
- **Gitignored.** \`.gitignore\` updated so the cookies file can never be committed accidentally.

## Applied to all 3 yt-dlp call sites

| File | Function | Purpose |
|------|----------|---------|
| \`backend/services/ingest.py\` | \`_download_youtube_sync\` | Actually downloads the audio |
| \`backend/services/cover_search.py\` | \`_yt_dlp_search\` | Finds cover candidates via \`ytsearch{N}:\` |
| \`backend/services/cover_search.py\` | \`_yt_dlp_extract_info\` | Probes a single URL's metadata |

All three invoke the shared \`apply_ytdlp_cookies()\` helper from the new \`backend/services/_ytdlp_utils.py\` — kept separate from \`ingest.py\` and \`cover_search.py\` to avoid circular imports (ingest already depends on cover_search).

## Ops: setting the secret

See the new \`docs/ytdlp-cookies.md\` for the 5-minute browser-export recipe. Once cookies are in hand:

\`\`\`bash
gh secret set OHSHEET_YTDLP_COOKIES --env qa   < ~/Downloads/youtube-cookies.txt
gh secret set OHSHEET_YTDLP_COOKIES --env prod < ~/Downloads/youtube-cookies.txt
gh workflow run \"Deploy to VM\" --ref qa   # or just push a commit
\`\`\`

Refresh cadence: roughly every 2-3 months, or immediately when users report the bot-check error.

## Test plan

- [ ] Merge + deploy to qa BEFORE the secret is set — confirm yt-dlp still runs anonymously (no regression)
- [ ] Set \`OHSHEET_YTDLP_COOKIES\` on qa environment, redeploy, confirm:
  - Container starts healthy
  - \`ssh deploy@34.169.16.93 \"cat ~/oh-sheet/youtube-cookies.txt | head -1\"\` shows Netscape header
  - \`sudo docker exec worker-ingest-1 ls -l /app/youtube-cookies.txt\` confirms mount
- [ ] Submit the failing URL again; error should not recur
- [ ] (Optional) Repeat for prod

## What this PR does NOT do

- Does not set the actual secret — that requires cookies exported from your own browser session (can't be done via CI)
- Does not change any existing test — no new tests added because the yt-dlp calls are already mocked at the test boundary (services patch \`_yt_dlp_search\`/\`_yt_dlp_extract_info\` directly). The cookies path only activates at production runtime.


jack0lantern:
Branch is stale and needs lint fixes

jack0lantern (dismissed):
Approved - merge when ci fixes are done.

  Minor observations                                                                                                                                                                                                                             
                                                                                                                                                                                                                                                 
  1. deploy.yml — long && chain in SSH command (deploy.yml:138): The atomic mv now chains 4 operations. If the cookies mv fails, .env and compose file are already renamed. Low risk since the failure mode is "re-run deploy," but worth noting 
  the chain is getting long.                                                                                                                                                                                                                     
  2. Import style inconsistency (intentional, fine): ingest.py imports apply_ytdlp_cookies at module top-level, while cover_search.py imports it locally inside each function. This is correct — cover_search.py uses local imports to keep
  yt_dlp out of the test import path, and the utility import follows that pattern. Just calling it out for future readers.
  3. No unit test for apply_ytdlp_cookies: The PR body explains why (yt-dlp calls are mocked at the service boundary). The function is simple enough that this is fine, but a 5-line test with a temp file would be easy to add if desired.      
  4. Cookies file scope: The bind-mount goes to orchestrator and worker-ingest but not worker-engrave or other services. This is correct since only ingest + cover_search use yt-dlp, but if a future service adds yt-dlp calls, the mount would
  need to be added to its compose entry too. The docs could mention this.

swifttarrow (approved):

PR #79, opened 2026-04-18
refactor(engrave): drop heuristic fallback, always engrave via ML service

swifttarrow:
## Summary

Retires the local music21 engraving stage. Audio/midi-upload jobs now always route through `oh-sheet-ml-pipeline`'s `/engrave` endpoint (paired PR: Oh-Sheet-Team/oh-sheet-ml-pipeline#14). There is no longer a fallback or a toggle — the ML service is the only engraver.

### Why

The local heuristic path (`backend/services/engrave.py`, music21 + LilyPond/MuseScore) produced poor scores and kept growing complexity as we patched around music21 quirks (OSMD compat, voice-collision chord tags, UUID part names, two-part brace encoding). Keeping it as a silent fallback only masked real problems with the ML service. With a trained seq2seq engraver online, the heuristic path is dead weight.

### What changed

**Pipeline**
- `backend/services/engrave.py` (~1.5K lines) and `backend/workers/engrave.py` deleted.
- `_render_midi_bytes` is the one piece the runner still needs (MIDI payload for the ML service); extracted to `backend/services/midi_render.py` as `render_midi_bytes`.
- `PipelineRunner` no longer branches on a toggle — audio_upload/midi_upload unconditionally POST MIDI to `OHSHEET_ENGRAVER_SERVICE_URL/engrave` and persist the returned MusicXML.
- title_lookup jobs that reach the engrave stage hard-fail with a clear error. Those are expected to resolve upstream via TuneChat; with no local fallback, silent failure at engrave would only hide TuneChat outages.

**Config / deploy**
- `OHSHEET_ENGRAVER_INFERENCE` retired. `OHSHEET_ENGRAVER_SERVICE_URL` (+ `…_TIMEOUT_SEC`) stay.
- `worker-engrave` removed from both `docker-compose.yml` and `docker-compose.prod.yml` — nothing for it to run.
- `deploy.yml` updated: drop the `vars.OHSHEET_ENGRAVER_INFERENCE` plumbing.

**API surface**
- `GET /v1/artifacts/{job_id}/pdf` returns **404** now. The ML engraver returns MusicXML only; PDF rendering is a client responsibility (the frontend already consumes MusicXML via OSMD).
- `POST /v1/stages/engrave` (stateless worker endpoint for external orchestrators) removed along with `EngraveService`.

**Tests**
- `tests/conftest.py` autoload-stubs `engrave_midi_via_ml_service` so pipeline tests don't need a live ML backend.
- Deleted: `test_engrave_quality.py` (~2K lines of music21 output assertions), `test_engrave_pdf_render.py`, `test_engraver_inference_toggle.py` (toggle is gone), `extract_musicxml.py` (debug helper for heuristic MusicXML).
- Updated: `test_celery_dispatch.py` / `test_jobs.py` / `test_artifacts.py` no longer assert `pdf_uri`; `test_ingest_youtube` asserts that a title_lookup job without TuneChat now fails (previously silently engraved).

### Test plan

- [x] `pytest` — 552 passed, 1 skipped
- [x] `ruff check backend/ tests/` — clean
- [x] `mypy backend/` — clean
- [ ] Re-provision `vars.OHSHEET_ENGRAVER_SERVICE_URL` in the GitHub env (remove the now-dead `vars.OHSHEET_ENGRAVER_INFERENCE`)
- [ ] Smoke: qa audio upload → `/engrave` returns real MusicXML from the seq2seq runner
- [ ] Smoke: qa midi upload → same
- [ ] Smoke: qa title_lookup with TuneChat enabled → TuneChat resolves upstream, engrave never runs
- [ ] Frontend check: `/v1/artifacts/{job}/pdf` 404 doesn't break the result screen (OSMD path only consumes `/musicxml`)


swifttarrow:
## Self-review

Reviewing only the three commits on this branch (8910003, fd737c3, a5b14e8) — the rest of the diff is inherited from qa.

### Blocking

**1. Orchestrator container won't see the new env vars.** `docker-compose.prod.yml` wires `OHSHEET_ENGRAVER_INFERENCE` / `OHSHEET_ENGRAVER_SERVICE_URL` through the VM `.env` via deploy.yml, but the `orchestrator:` service's `environment:` block in `docker-compose.prod.yml` never passes them into the container — unlike `OHSHEET_TUNECHAT_URL` / `OHSHEET_ANTHROPIC_API_KEY`, which are explicitly listed. Since the runner (which calls the ML service) runs inside the orchestrator container, the toggle will always read false in prod regardless of GitHub var values. My commit correctly removed the var from `worker-engrave` (which doesn't need it) but forgot to add it to `orchestrator`.

### Non-blocking

**2. No test coverage for error path.** `MLEngraverError` propagates out and fails the job — that's the intended behavior — but `test_engraver_inference_toggle.py` only exercises the happy path. One extra test where the fake client raises would document the contract.

**3. `score_dict → HumanizedPerformance` shell-builder duplicated.** The block at `backend/jobs/runner.py:454-483` duplicates the same shell construction in `backend/services/engrave.py:_engrave_sync` (≈L1363-1388). Either factor into a shared helper in `engrave.py` or accept the drift for now.

**4. No guard against `_STUB_MIDI` bytes.** The retired `_transcribe_via_ml_api` bailed when `_render_midi_bytes` fell back to the hand-rolled `_STUB_MIDI` (pretty_midi missing / empty performance). The new path ships those bytes to the service anyway. The service handles them gracefully (empty MusicXML), so not urgent — worth a comment if we keep it as-is.

### Plan

Addressing (1) and (2) in a follow-up commit. Leaving (3) and (4) for a later pass.

robin-raq (dismissed):
Thanks for picking this up! Verified the core contract: `title_lookup` jobs (TuneChat + cover_search) are correctly preserved — the `use_ml_engraver` gate excludes that source so `runner.py`'s TuneChat fast-path stays untouched. That's exactly the right shape.

Requesting three hardening additions before merge. None are design concerns — they're about making the current correctness more resilient to future edits.

### 1. Make the "errors now raise" behavior change visible in code

The PR body mentions:
> Errors now raise instead of silently falling back — the toggle itself is the off-switch.

That's a meaningful semantics change vs. the retired `OHSHEET_ML_API_URL` gate (which degraded to music21 on failure). Someone editing this block in 6 months won't know that from the diff alone. Please add a comment at the top of the `use_ml_engraver` branch in `backend/jobs/runner.py`:

```python
# Intentional behavior change vs. the retired OHSHEET_ML_API_URL gate:
# when this branch is taken, ML engraver failures PROPAGATE as job errors
# rather than degrading to the local music21 path. The boolean toggle
# itself is the off-switch — operators turn it off if the ML service is
# unhealthy, rather than relying on silent fallback.
```

### 2. Stub-deploy footgun — log a warning on first misuse

Paired PR #14 ships `/engrave` as a near-empty placeholder (header + one empty measure). If anyone flips `OHSHEET_ENGRAVER_INFERENCE=true` on qa/prod before the real seq2seq model is wired, every audio/midi upload produces a blank score with no alert.

Please add a one-shot health log in `ml_engraver_client.py` (or equivalent) that warns when the response looks like the stub payload:

```python
# ml_engraver_client.py — after a successful engrave call
if _looks_like_stub(musicxml_bytes):
    log.warning(
        "engraver_inference=true but response appears to be the ML "
        "pipeline stub placeholder. Confirm a real model is deployed."
    )
```

The `_looks_like_stub` check can be simple — response size threshold or a substring test against the header-only MusicXML skeleton PR #14 generates. Even a quick `len(musicxml_bytes) < 500` catches the empty-placeholder case.

### 3. Lock the title_lookup exclusion in a test

The exclusion of `title_lookup` from `use_ml_engraver` is the single most important invariant of this PR — it's what makes merging safe for the TuneChat integration. Please add a targeted test in `tests/test_engraver_inference_toggle.py`:

```python
@pytest.mark.parametrize("toggle", [True, False])
def test_title_lookup_never_routes_through_ml_engraver(toggle):
    # Regression guard for runner.py:use_ml_engraver gating.
    # title_lookup jobs (TuneChat + cover_search) must never hit the
    # ML engraver service regardless of OHSHEET_ENGRAVER_INFERENCE.
    settings.engraver_inference = toggle
    for source in ("title_lookup",):
        bundle = make_bundle(source=source)
        use_ml = _compute_use_ml_engraver(bundle)  # or assert inline
        assert use_ml is False, (
            f"title_lookup must bypass ML engraver (toggle={toggle}, "
            f"source={source}) — TuneChat's fast-path depends on this."
        )
```

If `use_ml_engraver` isn't factored out, refactoring it to a helper is a fine shape. Otherwise, assert the behavioral outcome (that the ML branch isn't taken) via mocks.

---

Happy to iterate on any of the above. Once these land I'll approve.

shoyu-ramen (commented):
Drive-by review from outside the paired-PR pair. The retirement rationale is persuasive — the music21 path was accreting patches faster than it was producing good output — and the inline-HTTP replacement is clean. A few things worth considering before merge.

### 1. `_looks_like_stub` warn-only re-introduces the failure mode this PR sets out to kill

The PR argues:

> Keeping it as a silent fallback only masked real problems with the ML service.

`ml_engraver_client._looks_like_stub` logs a warning and returns the bytes anyway. If the paired ML PR ships before a real seq2seq model is wired — or if the model regresses to the placeholder — every audio/midi upload will produce a near-empty MusicXML, `JobManager` reports success, and the user sees a blank score. That's the same masking shape, just relocated from the engraver to the client.

If the project's stance is "a non-real score is a failed job," I'd make it symmetric and raise `MLEngraverError` in `_looks_like_stub`. A log-only guard assumes someone is actively watching logs, which is exactly the safety net the PR argues against.

### 2. Silent localhost fallback in `docker-compose.prod.yml`

```yaml
OHSHEET_ENGRAVER_SERVICE_URL: ${OHSHEET_ENGRAVER_SERVICE_URL:-http://localhost:8080}
```

If `deploy.yml` ever forgets to populate this var (or the GitHub env `vars.OHSHEET_ENGRAVER_SERVICE_URL` gets unset per the test-plan note), the orchestrator boots happily and every engrave job fails against an unreachable `localhost:8080`. Given this is now the only engraver, I'd drop the default so the container refuses to start without an explicit value — or add a startup check in `backend/config.py` that raises if `engraver_service_url` is `http://localhost:*` outside of tests/dev.

### 3. No retry on transient ML service errors

A single `httpx.TimeoutException` or 5xx now fails the whole pipeline irrecoverably after ingest/transcribe/arrange/humanize have already spent compute. A light retry (e.g., 2× on timeout/5xx with brief backoff) inside `engrave_midi_via_ml_service` would improve robustness without reintroducing the fallback philosophy — different failure mode, not a fallback.

### 4. `pdf_uri = ""` as a "not produced" sentinel

`EngravedOutput(pdf_uri="")` plus `if not uri: 404` works, but empty-string-as-sentinel is subtle — callers writing `if result.pdf_uri is None` disagree with callers writing `if result.pdf_uri`. If the contract field is `str | None`, prefer `None`. If it's `str`, worth asking whether the field should remain on a MusicXML-only path at all.

### 5. `title_lookup` hard-fail is late

The `raise RuntimeError("title_lookup job reached the engrave stage without a TuneChat result...")` runs after four upstream stages have already executed. If TuneChat is disabled, the same check at `POST /v1/jobs` time would give users a clean 4xx immediately instead of burning the pipeline to fail at the end. Nit — the current behavior is correct, just expensive.

---

(1) and (2) are the ones I'd block on — both re-create the "silent bad output" failure mode the PR is specifically retiring the fallback to avoid. The rest are suggestions.

robin-raq:
Checked out this branch locally and ran the suite against the current tip (`278b561`). Sharing the evidence since it's relevant to the "is this safe to merge?" question.

## Test suite: green

```
555 passed, 0 failed, 112 warnings in 83.42s
TOTAL coverage: 81%
```

`test_ml_engraver_error_propagation.py` (the 2 tests documenting the invariant this PR relies on) passes clean. `test_refine_*` (46 tests from the inherited #64 changes) also green.

## Local stack startup — deferred

I wanted to `docker compose up` and smoke-test a real job against a stubbed `/engrave` endpoint, but hit two setup frictions worth flagging for anyone else pulling this branch:

1. **Base image requires `read:packages` scope on `gh` tokens.** `docker-compose.yml` targets `ghcr.io/oh-sheet-team/ohsheet-dev-base:latest` (private). Default `gh auth login` scopes are `repo, read:org, gist` — `docker pull` against ghcr gets denied. Fix: `gh auth refresh -s read:packages` before `echo $(gh auth token) | docker login ghcr.io -u <user> --password-stdin`.
2. **No local ML engraver.** Per this PR, `OHSHEET_ENGRAVER_SERVICE_URL` defaults to `http://localhost:8080`, which has nothing listening locally. Running an end-to-end job requires either spinning up `oh-sheet-ml-pipeline` alongside (paired PR #14) or pointing the env var at a deployed engraver. Per-run config, not a blocker for the PR itself.

Neither is a regression from this PR — (1) is inherited from the build-cache work, and (2) is the *intended* architectural shape. Worth a `docs/dev-setup.md` note for the next contributor, but doesn't affect merge decisions here.

robin-raq (dismissed):
Refreshed review — my previous CHANGES_REQUESTED referenced the since-retired `OHSHEET_ENGRAVER_INFERENCE` toggle, so I dismissed it and am re-submitting against the current code. Of my three original asks: #1 (behavior-change documentation) is close enough (the runner's hard-fail comment covers the same ground), #2 (stub detection) is implemented as `_looks_like_stub`, #3 (title_lookup test) is still unaddressed in the new shape.

## Blocking

### 1. The `title_lookup` hard-fail at the engrave stage has no regression test

`runner.py:429-434` is the single invariant that makes this PR safe for the TuneChat path:

```python
if bundle.metadata.source == "title_lookup":
    raise RuntimeError(
        "title_lookup job reached the engrave stage without "
        "a TuneChat result. ..."
    )
```

A future refactor that reorders the runner stages, moves the source check, or short-circuits the engrave branch could silently break this without a single test failing. `test_ml_engraver_error_propagation.py` covers audio_upload and midi_upload, not title_lookup. Please add a test that constructs a `title_lookup` bundle + a disabled TuneChat state and asserts `RuntimeError` (match `"title_lookup job reached the engrave stage"`) propagates.

### 2. Endorsing @shoyu-ramen's blocking points — both are correct

Both of his (1) and (2) align with the PR's own stated philosophy. Quoting his concerns because I want them on record as also coming from me, not just a drive-by observation:

- **`_looks_like_stub` warn-only re-introduces the silent-bad-output pattern this PR retires the fallback to avoid.** `ml_engraver_client.py:61-66` logs and returns the stub bytes — `JobManager` reports success, user sees a blank score. If the stance is "bad output is a failed job," the stub check should `raise MLEngraverError("stub response from engraver — confirm real model is deployed")`. Asymmetric otherwise.
- **`${OHSHEET_ENGRAVER_SERVICE_URL:-http://localhost:8080}` in `docker-compose.prod.yml:48`.** Given this is the only engraver path, a prod default pointing at an unreachable localhost is a deploy footgun — if `deploy.yml` ever drops the var, every job will fail against it. Either remove the default (container refuses to start on missing config) or add a `model_validator` in `backend/config.py` that rejects `localhost:*` when `OHSHEET_ENV != "dev"`.

### 3. `midi_render.py` has the mirror of shoyu's stub problem on the input side

`_STUB_MIDI` (32 bytes — empty track, EOT) is returned when:

- `pretty_midi` isn't installed (`midi_render.py:24-26`), or
- The humanized performance ends up with zero notes (`midi_render.py:82-83`)

Sending a stub MIDI to the ML engraver will get back stub MusicXML, same silent-blank-score failure shape. Two cases:

- **pretty_midi not installed**: this is a deploy configuration error, should fail the pipeline loudly (raise), not degrade silently. `pretty_midi` is a top-level dep in `pyproject.toml` — a missing import means someone installed a broken image.
- **Zero-note performance**: probably indicates a transcription regression or a silent audio file. Either way, a blank engrave is not a useful output. Raise with a clear message so the failure surfaces to the user/operator rather than producing a blank PDF-less result.

At minimum, log a warning with the same visibility as `_looks_like_stub` does on the response side — or better, raise symmetrically with whatever the project's choice on shoyu's #1 becomes.

## Non-blocking

- `ml_engraver_client.py:47` uses `response.content` (bytes), which is fine. Worth confirming the ML service sets `Content-Type: application/vnd.recordare.musicxml+xml` so the frontend's OSMD-only path has the right MIME for its fetch. Not a blocker but trivial to verify.
- `runner.py:451-454` lazy-imports `render_midi_bytes` / `engrave_midi_via_ml_service` inside the engrave branch. Fine for import-cycle avoidance; module-level imports would be slightly clearer if no cycle exists. Won't block on this.

## Context

Verified the core contract still holds against the current code. Full test suite (555 passed, 81% coverage) runs green on this branch — see my earlier comment with the local-run notes.

Once (1), shoyu's two items, and (3) are resolved, I'll approve. Happy to iterate on the specific shape of any of them.

robin-raq (approved):
Approving — all four blocking concerns from my (now-dismissed) CHANGES_REQUESTED are resolved, with implementation quality that went well beyond the minimum ask.

## How each concern was addressed

**1. Stub response raises (endorsed from @shoyu-ramen).** `MLEngraverStub(MLEngraverError, retryable=False)` replaces the warn-and-return. The exception message explicitly references the failure mode (*"service is likely running the in-tree placeholder rather than a real model. Refusing to surface a blank score."*). Symmetric with the PR's stated philosophy.

**2. Prod compose localhost default (endorsed from @shoyu-ramen).** `${OHSHEET_ENGRAVER_SERVICE_URL:?OHSHEET_ENGRAVER_SERVICE_URL must be set (no fallback engraver in prod)}` — docker-compose refuses to start with a clear error if the var is missing. Preferable to "drop the default" because the failure message is explicit about *why*.

**3. `title_lookup` hard-fail test.** Two new tests in `test_ml_engraver_error_propagation.py` — `test_title_lookup_hard_fails_when_tunechat_disabled` (the case I asked for) plus `test_title_lookup_hard_fails_when_tunechat_returns_none` (the defense-in-depth case I hadn't considered: TuneChat enabled but returning None mid-pipeline). Second test is a nice catch.

**4. `midi_render` stub fallback.** `_STUB_MIDI` deleted entirely. `MidiRenderError` raises on both missing `pretty_midi` (deploy misconfig) and empty-notes (likely transcription regression). The class docstring explicitly references symmetry with the response-side fix, which is the right framing — asymmetric silent-failure policies were the whole problem.

## Bonus improvements worth flagging

- **Exception hierarchy with `retryable` class attribute** — clean separation between transient (`MLEngraverTimeout`, `MLEngraverTransportError`, `MLEngraverUpstreamError`) and deterministic (`MLEngraverClientError`, `MLEngraverStub`) failures. Retry classification via `exc.retryable` instead of string-matching on error text is more maintainable (the docstring calls this out explicitly).
- **Retry loop with exponential backoff** (3 attempts, 0.5s base) — addresses @shoyu-ramen's non-blocking point about transient failures. `_post_once` cleanly separated from retry orchestration. Comments correctly distinguish retry-within-service from fallback.
- **11 new test cases** across `test_midi_render.py` (4) and `test_ml_engraver_client.py` (7). `test_4xx_fails_without_retry` and `test_stub_payload_does_not_retry` are the ones I'd have asked for to verify the retryable-classification design is actually wired correctly — glad they're there.

## Test run

- Targeted test files: **15/15 passing** (new + error-propagation)
- Full suite: **568/569** — one environment-dependent failure (`test_youtube_url_job_rejected_at_creation_without_tunechat`) passes with `OHSHEET_TUNECHAT_ENABLED=false` and fails with it set to true. The test doesn't monkey-patch the setting, so it depends on the default config. Not a PR #79 regression — this test existed before the PR and wasn't touched. Pre-existing flakiness + the fact that the failing path hits real YouTube (`[download] 100.0% of 3.27MiB` shows up in the captured output) is worth a separate test-hygiene ticket.

Nice work on the follow-up pass — the exception hierarchy in particular is a cleaner design than "retry these specific error strings" would have been.

PR #81, opened 2026-04-18
Revert PR #79 on qa: missing OHSHEET_ENGRAVER_SERVICE_URL blocks deploys

robin-raq:
## What

Reverts #79 on qa. Restores the previous commit (#80's merge) as qa tip.

## Why

PR #79 merged without provisioning \`OHSHEET_ENGRAVER_SERVICE_URL\` on the qa GitHub environment. Deploy workflow fails at \`docker compose up\` with:

\`\`\`
error while interpolating services.orchestrator.environment.OHSHEET_ENGRAVER_SERVICE_URL: required variable OHSHEET_ENGRAVER_SERVICE_URL is missing a value: OHSHEET_ENGRAVER_SERVICE_URL must be set (no fallback engraver in prod)
\`\`\`

This blocks **all** deploys to qa, including PR #80 (yt-dlp cookies fix) whose secret needs to propagate to the VM so YouTube ingest works at all.

## Scope

- qa only — main/prod not touched
- Title-lookup / YouTube / TuneChat path unaffected (ML engraver gate excludes \`source="title_lookup"\`)
- @swifttarrow please re-land after coordinating env-var provisioning on qa

## Verification

- \`git log origin/qa --oneline\` after merge should show this revert on top
- Next qa deploy should succeed and propagate \`OHSHEET_YTDLP_COOKIES\`


robin-raq:
Going with option 2 instead — set OHSHEET_ENGRAVER_SERVICE_URL=http://disabled on qa env var. Title-lookup path doesn't touch ML engraver, and audio/midi uploads will fail at runtime instead of blocking deploys. PR #79's author can still provision the real URL when the service is ready.

PR #82, opened 2026-04-18
perf(ci): add Docker layer caching to speed up deploys

lramosve:
## Summary

Replace raw `docker build` + `docker push` commands with `docker/build-push-action@v6` + GitHub Actions layer cache. Each image gets its own cache scope.

### Expected impact

The app image build is the bottleneck (~5 min of the ~12 min deploy). The heavy layers that rarely change:
- `pip install .[pop2piano]` — torch + transformers (~800 MB, changes only when deps update)
- Flutter web build (~60s, changes only when frontend code changes)
- `apt-get install lilypond ffmpeg` (~8s, almost never changes)

With `cache-from: type=gha`, these layers are pulled from GitHub's cache on cache-hit deploys. Expected build time: **~1-2 min** (down from ~5 min) for code-only changes.

### Changes

- Added `docker/setup-buildx-action@v3` (required for GHA cache backend)
- Split the single `run:` block into three `docker/build-push-action@v6` steps (app, decomposer, assembler)
- Each step uses `cache-from: type=gha,scope=<name>` / `cache-to: type=gha,scope=<name>,mode=max`

### First deploy

The first deploy after merge will be a cold cache (no speedup). Every subsequent deploy will benefit.

## Test plan

- [ ] Merge and verify QA deploy succeeds (first run populates cache)
- [ ] Push a code-only change and verify the second deploy is faster (~5-7 min total vs ~12 min)

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -52,35 +52,56 @@ jobs:
       - name: Authorize Docker push
         run: gcloud auth configure-docker ${{ env.REGION }}-docker.pkg.dev --quiet
 
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+
       - name: Extract version
         id: version
         run: |
           VERSION=$(grep -m1 '^version' pyproject.toml | sed 's/version *= *"\(.*\)"/\1/')
           echo "version=$VERSION" >> "$GITHUB_OUTPUT"
 
-      - name: Build and push container images
-        env:
-          # TUNECHAT_URL is baked into the Flutter JS bundle at build time
-          # via --dart-define (see Dockerfile flutter-build stage).
-          # Pulling through env: is the safe pattern — inline expressions
-          # in a run: shell line is a workflow-injection smell.
-          TUNECHAT_URL: ${{ vars.OHSHEET_TUNECHAT_URL }}
-        run: |
-          docker build \
-            --build-arg APP_VERSION=${{ steps.version.outputs.version }} \
-            --build-arg TUNECHAT_URL="$TUNECHAT_URL" \
-            -t ${{ env.IMAGE }}:${{ github.sha }} \
-            -t ${{ env.IMAGE }}:latest .
-          docker push ${{ env.IMAGE }}:${{ github.sha }}
-          docker push ${{ env.IMAGE }}:latest
+      # Build all three images with GitHub Actions layer cache.
+      # The app image is the big win (~5 min uncached → ~1 min cached)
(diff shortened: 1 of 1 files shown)


swifttarrow (approved):

jack0lantern (commented):
## Review: Docker Layer Caching for CI Deploys

Single file change (`.github/workflows/deploy.yml`, +44/-23). Clean, well-scoped PR. Two issues worth addressing before merge:

---

### Critical: Buildx credential handoff may break `push`

`gcloud auth configure-docker` writes a `credHelpers` entry to `~/.docker/config.json`. The old code used `docker push` directly in a shell step, which reads that file. The new code uses `docker/build-push-action@v6` with a buildx builder spun up by `docker/setup-buildx-action@v3`.

The problem: `setup-buildx-action` creates a `docker-container` driver by default — that builder runs inside a container and **does not see the host's gcloud credential helper**. This can produce a silent 401 on push (documented in docker/buildx#819).

**Fix:** Add `docker/login-action@v3` between setup-buildx and the build steps:

```yaml
- name: Authenticate to Google Cloud
  id: auth                          # ← add id
  uses: google-github-actions/auth@v2
  ...

- name: Log in to Artifact Registry
  uses: docker/login-action@v3
  with:
    registry: ${{ env.REGION }}-docker.pkg.dev
    username: oauth2accesstoken
    password: ${{ steps.auth.outputs.auth_token }}
```

This is the pattern Google's own docs endorse for the buildx path.

---

### Important: GHA cache budget pressure with `mode=max`

GitHub Actions cache is 10 GB per repo. The app image alone has a ~1.6 GB torch layer, and `mode=max` exports **every** intermediate layer. Three images with `mode=max` will approach or exceed the limit, causing eviction of the very layers you want cached most.

**Fix:** Use `mode=min` for decomposer and assembler — they have no heavy layers, so `mode=max` adds cache cost with no meaningful benefit:

```yaml
# decomposer & assembler steps:
cache-to: type=gha,scope=decomposer,mode=min
cache-to: type=gha,scope=assembler,mode=min
```

Keep `mode=max` only for the app image where it matters.

---

### Everything else looks correct

- Build args, tags, contexts, and Dockerfile paths all map correctly from the original shell commands
- `TUNECHAT_URL` in `build-args:` is safe — it's a structured YAML input to the action (not shell-interpolated), and `vars.*` are repo variables, not untrusted input
- Sequential build order is correct for a single-runner deploy
- Action version pinning is consistent with the rest of the workflow
- Permissions block (`id-token: write`) covers both WIF auth and GHA cache

PR #83, opened 2026-04-19
Frontend-v2 SPA + cover_search title cleanup

robin-raq:
## Summary

Three self-contained commits landing in one branch:

1. **`feat(cover_search): allowlist Peter Bence + @PianoTutorialEasy`** — policy change only, +3 LOC. Peter Bence's piano covers were scoring 50 (below the 60 threshold) because he was missing the +50 allowlist bonus despite ~3M subs of high-quality arrangements. @PianoTutorialEasy added to the easy-tier channels from the same audit.

2. **`feat(ingest): return extracted title/artist from cover_search`** — internal signature change to \`_maybe_swap_for_cover_sync\` so the caller can thread the ORIGINAL URL's cleaned title/artist (split on \`" - "\` / \`" – "\` / \`" | "\`, trailing parens stripped) into the engraved sheet's metadata. Result: the PDF header reads \"Beat It\" / \"Michael Jackson\" instead of the uploader's raw video title like \"BEAT IT - Michael Jackson x Peter Bence (Piano Cover)\".

3. **`feat: frontend-v2 vanilla-JS SPA`** — new \`frontend-v2/\` directory parallel to the existing Flutter frontend. **Not yet wired into the Docker build** — Flutter still serves qa after this merges. A follow-up PR will swap the static-file mount in \`backend/main.py\` and add the npm build step to the \`Dockerfile\`.

## Why replace Flutter (eventually)

- ~6 MB first-paint (CanvasKit WASM + Dart runtime + app) vs ~100 KB vanilla-JS bundle → 3-5 s cold load on mid-range mobile drops to ~300 ms
- CanvasKit rendering is battery-hungry on mobile
- DOM-based rendering means accessibility, SEO, and devtools debugging all work normally
- The app is web-only — Flutter's cross-platform value isn't redeemed

## Architecture of frontend-v2

- \`src/api.js\` — fetch wrappers for /v1/uploads, /v1/jobs, WS  (13 tests)
- \`src/state.js\` — pure reducer + tiny pub/sub store  (19 tests)
- \`src/views.js\` — phase renderers for the morphing card  (30 tests)
- \`src/main.js\` — bootstrap wiring + mascot swap + WS lifecycle
- \`src/style.css\` — cream palette, card morph, mobile responsive
- \`index.html\` — app shell with mascot slot + card container

Parallel TDD across api/state/views modules with a shared \`src/SPEC.md\` contract so the modules could be built concurrently by separate agents and integrated cleanly.

## Test plan

[...]

PR #85, opened 2026-04-19
fix(frontend-v2): WS lifecycle + drop dead title source + TUNECHAT_TIMEOUT plumbing

robin-raq:
## Summary

Two commits from this morning's PR #83 code review + a demo-day blocker.

### Commit 1: \`fix(frontend-v2): WS lifecycle + drop dead title source branches\`

Addresses three 🟡 issues from the PR #83 review:

1. **Dead code: \"title\" source removed.** The segmented picker only exposes audio/midi/youtube after #83, so the \`source === \"title\"\` branches in views.js (idleBody + submit) and main.js (onSubmit) were unreachable. The title FIELD as an optional input on audio/midi uploads is preserved — only title-as-source-type is gone.

2. **WS subscribeToJob now handles onclose + onerror.** api.js synthesizes a retryable \`{type: \"job_failed\"}\` event when the socket drops before a terminal event. Without this, a network blip / proxy timeout / backend restart left the UI stuck in the last \"working:...\" phase forever. Dedupe via a \`closed\` flag so Chrome's onerror-then-onclose sequence doesn't double-emit; Firefox's onclose-only path also works. Unsubscribe + terminal events mark the flow as intentional so no spurious error fires during normal cleanup.

3. **onSourceChange now closes the in-flight WS.** main.js' onSourceChange reset the phase to idle but left the old unsubscribeWs alive — a late job_succeeded from the previous job would clobber the fresh idle state with a stale Complete phase.

### Commit 2: \`feat(deploy): plumb OHSHEET_TUNECHAT_TIMEOUT_SEC through to containers\`

Demo-day blocker surfaced this morning while testing long tracks locally:

- TuneChat takes 6–10 min to transcribe a 10-min audio file
- oh-sheet's httpx client timeout defaults to **300 s (5 min)**
- Past 5 min, the client gives up, runner.py falls through to local engrave, which hard-fails for \`title_lookup\` jobs (PR #79's \"no fallback engraver\")
- User-facing error: \"title_lookup job reached the engrave stage without a TuneChat result\" — looks like TuneChat is broken, but it's actually a 5-min client impatience

Plumbs \`OHSHEET_TUNECHAT_TIMEOUT_SEC\` through three layers:

1. **GitHub env-scoped variable** — set to \`900\` on qa environment (done via \`gh variable set\` out-of-band)
2. **\`.github/workflows/deploy.yml\`** — reads the var, writes to the VM's \`.env\`
3. **\`docker-compose.prod.yml\`** — both orchestrator + worker-ingest read from the host \`.env\` with a 300 s fallback

Fallback to 300 s in compose means the plumbing is **safe to deploy without the variable set** — behaviour matches today's default. Once set, the deploy propagates the higher timeout.

### Verified locally

- 10-min WAV (111 MB, 48kHz stereo) uploaded through TuneChat (Multer cap also raised — separate TuneChat-repo change), transcribed in 6m13s, job reports \`succeeded\` with \`tunechat_job_id\` populated
- 65/65 frontend-v2 tests passing (+4 new WS lifecycle cases)

## Test plan

- [x] \`cd frontend-v2 && npm test\` — 65/65 passing
- [x] Local e2e with a 10-min track → TuneChat fast-path succeeds end-to-end (was timing out at 300s before this PR)
- [ ] Post-merge: qa deploys, smoke-test a long-track URL → expect \`job_succeeded\` with \`tunechat_job_id\` inside 15 min

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -115,6 +115,11 @@ jobs:
           OHSHEET_TUNECHAT_ENABLED: ${{ vars.OHSHEET_TUNECHAT_ENABLED }}
           OHSHEET_TUNECHAT_URL: ${{ vars.OHSHEET_TUNECHAT_URL }}
           OHSHEET_TUNECHAT_API_KEY: ${{ secrets.OHSHEET_TUNECHAT_API_KEY }}
+          # Long YouTube tracks (up to ~20 min) need a client timeout
+          # higher than the default 300 s — TuneChat can take 6-10 min
+          # to transcribe a 10-min audio file. 900 s is recommended for
+          # qa / prod. Unset falls back to 300 s (backend/config.py).
+          OHSHEET_TUNECHAT_TIMEOUT_SEC: ${{ vars.OHSHEET_TUNECHAT_TIMEOUT_SEC }}
           # Anthropic API — used by worker-refine for LLM metadata enrichment.
           OHSHEET_ANTHROPIC_API_KEY: ${{ secrets.OHSHEET_ANTHROPIC_API_KEY }}
           # Base URL of oh-sheet-ml-pipeline engraver service. The engrave
@@ -151,6 +156,7 @@ jobs:
           echo "OHSHEET_TUNECHAT_ENABLED=${OHSHEET_TUNECHAT_ENABLED}" >> /tmp/.env
           echo "OHSHEET_TUNECHAT_URL=${OHSHEET_TUNECHAT_URL}" >> /tmp/.env
           echo "OHSHEET_TUNECHAT_API_KEY=${OHSHEET_TUNECHAT_API_KEY}" >> /tmp/.env
+          echo "OHSHEET_TUNECHAT_TIMEOUT_SEC=${OHSHEET_TUNECHAT_TIMEOUT_SEC}" >> /tmp/.env
           # Anthropic — worker-refine LLM metadata enrichment.
           echo "OHSHEET_ANTHROPIC_API_KEY=${OHSHEET_ANTHROPIC_API_KEY}" >> /tmp/.env
           echo "OHSHEET_ENGRAVER_SERVICE_URL=${OHSHEET_ENGRAVER_SERVICE_URL}" >> /tmp/.env
--- a/docker-compose.prod.yml
+++ b/docker-compose.prod.yml
@@ -43,6 +43,11 @@ services:
       OHSHEET_TUNECHAT_ENABLED: ${OHSHEET_TUNECHAT_ENABLED:-false}
       OHSHEET_TUNECHAT_URL: ${OHSHEET_TUNECHAT_URL:-}
       OHSHEET_TUNECHAT_API_KEY: ${OHSHEET_TUNECHAT_API_KEY:-}
+      # Client-side timeout for the httpx POST to TuneChat. 300 s default
+      # is short for 10+ min tracks — tuned higher on qa/prod via the
+      # deploy.yml → .env pipeline. Default falls through to
+      # backend/config.py's 300 s if the env var is unset.
+      OHSHEET_TUNECHAT_TIMEOUT_SEC: ${OHSHEET_TUNECHAT_TIMEOUT_SEC:-300}
       # Engraver service URL — backend.jobs.runner (living in this container)
       # always routes audio/midi engrave jobs to this URL. No local fallback.
       # REQUIRED: ``${VAR:?msg}`` makes docker compose refuse to start when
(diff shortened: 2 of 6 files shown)


swifttarrow:
## Code Review — PR #85

### Overview
Two narrowly-scoped fixes stacked in one PR:
1. **Frontend-v2:** close the WS on `onSourceChange`, synthesize a retryable `job_failed` on abnormal close/error, and remove dead `source === "title"` branches in `views.js` + `main.js` (post PR #83's picker simplification).
2. **Deploy:** plumb `OHSHEET_TUNECHAT_TIMEOUT_SEC` from env-scoped GitHub variable → `.env` → both compose services, with a 300 s fallback so this is safe to deploy even before the variable is set.

Overall solid. Commit messages explain the *why* well, tests cover the new WS branches, and the docker-compose fallback makes the deploy change drop-in safe.

### What I like
- **Dedupe via `closed` flag** in `api.js:114–128` correctly handles the Chrome `onerror → onclose` sequence and the Firefox `onclose`-only path without double-emitting. Tested explicitly in `api.test.js:283–294`.
- **`unsubscribed` is set before `ws.close()`** (`api.js:150`), so the subsequent onclose correctly sees intentional cleanup. Without that ordering, unsubscribe would itself synthesize a spurious error.
- **Compose fallback** `${OHSHEET_TUNECHAT_TIMEOUT_SEC:-300}` means an empty/unset GitHub var still yields a valid integer — pydantic's `tunechat_timeout_sec: int` in `backend/config.py:598` wouldn't handle an empty string, so the fallback here is load-bearing.
- **Terminal-flag gate** on the synthesis check: a real `job_succeeded` followed by a normal close doesn't double-emit a failure. Good.
- **`onSourceChange` WS cleanup** (`main.js:78`) closes the scenario where a stale late `job_succeeded` clobbers a fresh idle.

### Issues / suggestions

#### 🟡 Minor — late-message window still open during source tab switch
`onSourceChange` now unsubscribes, but the WS `onmessage` handler doesn't check the `unsubscribed` flag. In the micro-window between `ws.close()` being called and the browser actually tearing down the socket, a late frame could still fire `onEvent` and overwrite the fresh idle phase via the reducer. In practice browser `close()` cancels pending tasks synchronously so this is usually safe, but cheap to harden:

```js
ws.onmessage = (ev) => {
  if (unsubscribed) return;
  // ...existing parse + dispatch
};
```

Same thing guards `onSubmit`'s terminal-event race if the user hits retry mid-message.

#### 🟡 Minor — synthesized `job_failed` event shape
`api.js:97–104` builds the synthesized event with `stage: null, progress: null`. The reducer (`state.js:68`) only reads `event.message`, so this works today, but any future reducer change that dereferences `event.stage.foo` would silently crash on the synthesized path. Consider either dropping the nulls (reducer uses `??`) or adding a brief `// intentionally null — not a pipeline-emitted event` note.

#### 🟡 Minor — `OHSHEET_TUNECHAT_TIMEOUT_SEC` operational coordination
The PR description mentions the variable was set to `900` on qa via `gh variable set` out-of-band. Two follow-ups worth flagging:
- **Prod env** also needs the var set (or it'll silently deploy with 300 s and repeat the symptom).
- Consider a note in the deploy runbook / `.env.example` so the next person doesn't have to reverse-engineer why long-track jobs intermittently fail.

#### 🟢 Nit — views.js fallback comment
`views.js:223–225` now uses `else { /* youtube */ }` as a catch-all. The comment is clear, but if the segmented picker ever grows a fourth option without updating this branch, silent fallback to youtube will be surprising. An explicit `source === "youtube"` check with an assertion/log for the unknown case would be more defensive. Low priority.

### Risks
- **WS reconnection semantics change:** users on flaky networks will now see \"Connection lost — try again\" instead of a hung spinner. That's strictly better UX, but worth watching the error rate in qa post-merge — if it's noisy, consider a short grace-period retry before surfacing the error.
- **TuneChat timeout upshift:** raising the client timeout to 900 s means a wedged TuneChat pins a Celery worker for 15 min. Per the comment in `backend/config.py:593`, worker concurrency should be sized accordingly on qa/prod — worth confirming the current concurrency has headroom.

### Tests
- 4 new WS lifecycle tests in `api.test.js` cover the four relevant paths (abnormal close, post-terminal close, unsubscribe, onerror dedupe). Good coverage.
- No test for `onSourceChange` closing the inflight WS — that change is in `main.js`, which is less testable as structured, so probably fine to skip, but a smoke note on the manual test would be nice.

### Verdict
LGTM with the minor hardening suggestions above. The compose fallback makes commit 2 safe to merge independently; commit 1's WS synthesis is a clear UX win. Non-blocking on the late-message-window suggestion.

swifttarrow (approved):

jack0lantern (approved):
LGTM. WS lifecycle fix is solid — the three-flag state machine handles the Chrome/Firefox close/error differences cleanly, and the synthesized event with `data.synthesized: true` is a nice touch for downstream disambiguation. Dead code removal and timeout plumbing are straightforward. Good test coverage on the edge cases.

PR #87, opened 2026-04-19
fix(frontend-v2): drop late WS frames after unsubscribe

robin-raq:
## Summary
- Guards \`ws.onmessage\` on the existing \`unsubscribed\` flag so late frames queued between \`ws.close()\` and browser teardown can't clobber a freshly-set phase.
- Adds a comment on the synthesized \`job_failed\` event's null \`stage\`/\`progress\` fields so future reducer changes don't silently deref them.

## Follow-up to
PR #85 review — the reviewer flagged the late-message window as a minor hardening nit. This lands it.

## Test plan
- [x] \`npm test\` — 66/66 passing (new test: unsubscribe → fire message → assert no dispatch)
- [ ] Manual: switch source tabs mid-job on qa, confirm idle stays idle

### DIFF
--- a/frontend-v2/src/api.js
+++ b/frontend-v2/src/api.js
@@ -100,6 +100,12 @@ export function subscribeToJob(jobId, onEvent) {
   let unsubscribed = false;
 
   ws.onmessage = (ev) => {
+    // Late-frame guard: between unsubscribe() calling ws.close() and the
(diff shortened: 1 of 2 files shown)


lramosve (approved):
Looks good, ship it!!

PR #89, opened 2026-04-19
feat(frontend-v2): widen results card + fullscreen button on TuneChat iframe

robin-raq:
## Summary
Two complementary fixes so the embedded TuneChat iframe on the results page renders at a readable width on desktop:

1. **Widen the card on complete** — CSS-only, hooks off the existing \`body[data-phase="complete"]\`. Card jumps from 560px to \`min(1200px, 90vw)\`; iframe height from 80vh→85vh. Reading-width territory.
2. **Fullscreen overlay button** — top-right corner of the iframe (YouTube / Vimeo / Figma convention). Uses the browser Fullscreen API (Escape-to-exit, iOS Safari chrome hiding, SR mode announcements for free). Hover-reveal on desktop, always-visible + 44×44 on touch.

## Screenshots (desktop)
Before: iframe pinned to 560px card → OSMD staves wrap at 2-3 measures/line.
After: iframe renders at ~1200px → matches the \`/embed\` URL's layout.

## Test plan
- [x] \`npm test\` — 68/68 (3 new: renders fullscreen button, enters FS on click, exits FS when already FS)
- [x] \`npm run build\` — CSS 10.73 kB (up from 10.00 kB), JS 12.04 kB
- [ ] Manual: qa job complete → card widens, iframe fills → fullscreen button hover-reveals → click → iframe goes fullscreen → Esc exits
- [ ] Manual mobile: fullscreen button always visible, 44×44 hit target

### DIFF
--- a/frontend-v2/src/style.css
+++ b/frontend-v2/src/style.css
@@ -223,11 +223,49 @@ body[data-phase="submitting"] .tagline { display: none; }
   height: min(70vh, 720px); background: var(--surface);
   border: 2px solid var(--stroke); border-radius: 20px;
   margin-bottom: 16px; overflow: hidden;
+  /* position:relative anchors the fullscreen overlay button. */
+  position: relative;
 }
 .iframe-stub iframe {
   width: 100%; height: 100%; border: 0; display: block;
 }
 .iframe-stub.tunechat { height: min(80vh, 760px); }
+
+/* Fullscreen overlay button — top-right corner of the iframe wrap,
+   matching YouTube / Vimeo / Figma convention. Hover-reveals on
+   desktop so the chrome stays minimal; always visible on touch. */
+.fullscreen-btn {
+  position: absolute;
+  top: 12px; right: 12px;
+  width: 36px; height: 36px;
+  display: inline-flex; align-items: center; justify-content: center;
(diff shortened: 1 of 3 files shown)


jack0lantern (approved):
Clean, well-scoped PR — CSS-only card widening is solid, a11y story is good (aria-label, focus-visible, 44×44 touch targets), and tests cover the key paths. Approving, with a few notes for follow-up:

**Fullscreen targets the iframe, not the wrapper** — `frame.requestFullscreen()` puts the iframe element fullscreen, but the overlay button is a sibling, not a descendant — so it disappears once fullscreen activates and the user can only exit via Escape. The exit code path in the click handler is effectively unreachable. Consider calling `requestFullscreen()` on the `.iframe-stub` wrapper instead so both the iframe and the button stay visible. Minor since Escape works, but worth a follow-up.

**Test cleanup** — the two fullscreen tests mock `document.fullscreenElement` via `Object.defineProperty` but never restore it, so the second test's mock leaks to subsequent suites. An `afterEach` restore would tighten this up.

**Nit:** `position: relative` lands on both `.iframe-stub` and `.inline-render` via the combined selector — harmless but could be scoped to `.iframe-stub` only since the comment says it's for the fullscreen button.

PR #93, opened 2026-04-20
feat(demo): YouTube-only landing + fix PDF/MusicXML/MIDI downloads via TuneChat proxy

robin-raq:
## Summary
Two demo-day blockers in one PR (independent but both urgent for tomorrow).

### 1. \`feat(frontend-v2)\`: YouTube-only landing page
Audio + MIDI source options aren't plumbed through the TuneChat fast-path, so the picker shouldn't advertise them. SOURCES reduced to \`[youtube]\`; segmented picker auto-hides when SOURCES.length <= 1. Backend endpoints + upload flow untouched — restoring post-demo is a 2-line revert.

### 2. \`fix(artifacts)\`: PDF/MusicXML/MIDI downloads were 404ing
**Root cause:** On TuneChat-fast-path jobs (the demo path), \`runner.py\` wrote \`EngravedOutput\` with empty local-blob URIs because the local engraver is skipped. The artifacts endpoint then returned 404 for every download. Meanwhile TuneChat's transcribe response *does* include the artifact URLs (\`fullMidiUrl\`, \`fullMusicxmlUrl\`, \`fullPdfUrl\`) — Oh Sheet was just throwing them away.

**Fix:** Capture the URLs in \`TuneChatResult\`, add optional fields to \`EngravedOutput\`, pass them through in \`runner.py\`, and in \`artifacts.py\` proxy the fetch via \`httpx\` when a tunechat_*_url is set. Falls back to the local blob-store path for audio/midi upload pipelines.

**Why proxy instead of 302 redirect:** the \`<a download>\` attribute is unreliable across cross-origin redirects (PDF opens inline, MIDI doesn't auto-download). Proxying lets Oh Sheet set \`Content-Disposition: attachment\` server-side so browsers always download.

## Test plan
- [x] \`npm test\` — 61/67 passing + 6 skipped (audio/midi describe blocks marked \`.skip\` with restore notes)
- [x] \`npm run build\` — green
- [x] \`.venv/bin/pytest tests/\` — 504/504 + 12 skipped (pre-existing transcribe_stems flake on qa, unrelated)
- [x] 4 new artifacts tests covering proxy-musicxml / proxy-pdf / local-fallback / upstream-502
- [ ] Manual on qa: click PDF/MusicXML/MIDI chips on a completed YouTube job → files download with correct names, not 404

### DIFF
--- a/backend/api/routes/artifacts.py
+++ b/backend/api/routes/artifacts.py
@@ -37,11 +37,12 @@ ArtifactKind = Literal["pdf", "musicxml", "midi", "transcription_midi"]
 # when the TuneChat field is empty (e.g. audio_upload / midi_upload paths
 # that never touch TuneChat). ``transcription_midi`` has no TuneChat
 # counterpart — it's the pre-engrave MIDI from Oh Sheet's own pipeline.
+_MUSICXML_MIME = "application/vnd.recordare.musicxml+xml"
 _KIND_INFO: dict[str, tuple[str, str | None, str, str]] = {
-    "pdf":                ("pdf_uri",                "tunechat_pdf_url",      "application/pdf",                       "sheet.pdf"),
-    "musicxml":           ("musicxml_uri",           "tunechat_musicxml_url", "application/vnd.recordare.musicxml+xml", "score.musicxml"),
-    "midi":               ("humanized_midi_uri",     "tunechat_midi_url",     "audio/midi",                            "humanized.mid"),
-    "transcription_midi": ("transcription_midi_uri", None,                    "audio/midi",                            "transcription.mid"),
+    "pdf":                ("pdf_uri",                "tunechat_pdf_url",      "application/pdf", "sheet.pdf"),
+    "musicxml":           ("musicxml_uri",           "tunechat_musicxml_url", _MUSICXML_MIME,    "score.musicxml"),
+    "midi":               ("humanized_midi_uri",     "tunechat_midi_url",     "audio/midi",      "humanized.mid"),
+    "transcription_midi": ("transcription_midi_uri", None,                    "audio/midi",      "transcription.mid"),
 }
 
 # httpx timeout for the TuneChat proxy fetch. Short because the file
--- a/backend/api/routes/artifacts.py
+++ b/backend/api/routes/artifacts.py
@@ -4,11 +4,20 @@ The pipeline contracts return ``file://`` URIs in ``EngravedOutput`` so the
 worker layer can stay storage-agnostic. Browsers and mobile apps can't fetch
 those URIs directly, so this route resolves the kind → URI for a finished
 job, reads the bytes via the BlobStore, and streams them back over HTTP.
+
+For the TuneChat-fast-path (title_lookup jobs routed through the external
+TuneChat engraver), artifacts live on TuneChat's server rather than in our
+blob store — ``EngravedOutput.tunechat_*_url`` holds the full https URL.
+This endpoint proxies those downloads so the client-facing URL shape
+(``/v1/artifacts/{job}/{kind}``) stays consistent regardless of which
+engraver produced the bytes.
 """
 from __future__ import annotations
 
+import logging
 from typing import Annotated, Literal
 
+import httpx
 from fastapi import APIRouter, Depends, HTTPException
 from fastapi.responses import Response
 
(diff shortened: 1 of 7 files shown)


lramosve (dismissed):
It all makes sense now (says Confucious LOL). Ship it!!

swifttarrow:
## Code Review — PR #93

### Overview
Two independent demo-blocker fixes bundled:
1. **Frontend**: `SOURCES` reduced to `[youtube]`; segmented picker auto-hides when `SOURCES.length <= 1`.
2. **Backend**: Capture TuneChat artifact URLs in `TuneChatResult` → `EngravedOutput`; `/v1/artifacts/{job}/{kind}` proxies the fetch via `httpx` with server-side `Content-Disposition: attachment`.

The proxy-over-redirect rationale in the PR body is correct and worth keeping in the docstring.

### Strengths
- **Clear WHY comments** — the `SOURCES` cut, the `segmented()` auto-hide guard, and the artifacts docstring all explain the demo-scope reasoning and the "2-line revert" restore path. Makes future-me's job easy.
- **Good test coverage on the new path** — 4 new tests cover proxy-musicxml, proxy-pdf, local-fallback, and upstream-502. The fallback test is the important one: it proves audio/midi upload pipelines still work.
- **Correct error taxonomy** — upstream failure → 502 (not 500). `httpx.HTTPError` base catches both `ConnectError` and `HTTPStatusError`.
- **`transcription_midi` handled safely** — `tunechat_attr=None` skips the URL check entirely; no risk of a missing TuneChat field breaking local-only artifacts.
- **Typed dict for `_KIND_INFO`** with explicit tuple shape is readable enough given only 4 entries.

### Issues & Suggestions

**1. Potential SSRF — URL is trusted blindly** (`backend/api/routes/artifacts.py:87`)
The proxy fetches whatever URL is in `tunechat_*_url`. Currently that's set only from TuneChat's API response, so the trust chain is "we trust TuneChat's hostname." Defense-in-depth: validate host against an allowlist (e.g. `settings.tunechat_base_url` host) before firing the `httpx` request. One-liner, makes the security story robust if TuneChat ever mirrors files from untrusted CDNs.

**2. Per-request `httpx.Client` instantiation** (`artifacts.py:84`)
Creating a new client per download is wasteful — no connection pooling, TLS handshake on every hit. For demo scale this is invisible, but a module-level `httpx.Client` (or a small `@lru_cache`) would be cleaner. Low priority.

**3. Entire response body loaded into memory** (`artifacts.py:89-92`)
`response.content` buffers the full artifact. Sheet-music PDFs are KB-sized so fine, but if TuneChat ever serves something larger this becomes a memory cliff. If you want to future-proof, `client.stream("GET", url)` + `StreamingResponse` is the drop-in upgrade. Not blocking.

**4. `inline` param is referenced but not visible in the diff** (`artifacts.py:78`)
The proxy branch correctly honors it, but worth a test asserting `?inline=true` produces `Content-Disposition: inline` on the TuneChat path. Easy to miss during a later refactor.

**5. Test helper reaches into DI internals** (`tests/test_artifacts.py:92`)
`client.app.dependency_overrides.get(get_job_manager, get_job_manager)()` is clever but fragile — if the test ever runs under a different DI pattern (e.g. `get_job_manager` becomes async), this silently breaks. Consider exposing a fixture that returns the live `JobManager` instance the app is using.

**6. Duplicated `FakeClient` across 3 tests**
The `__init__/__enter__/__exit__/get` scaffold is repeated. Extract a `_make_fake_httpx_client(body=..., raises=None)` helper — would make the 502 test a 2-liner. Minor.

**7. Frontend test `renderPhase — idle / title`** (`views.test.js:150`)
Sentinel changed from `.segmented` to `.body`. The comment correctly notes why, but `.body` is a generic class name — if idleBody() ever renames its wrapper, this test breaks silently. Consider a more specific sentinel like the presence of the `ytUrl`/YouTube input, consistent with the atomic-swap test you already updated.

**8. Log line exposes full TuneChat URL on failure** (`artifacts.py:92`)
Currently TuneChat URLs look like path-only `/pipeline/<id>/notation.mid` — safe. If TuneChat ever starts signing URLs with query-string tokens, this log becomes a credential leak. Low risk today; worth a note in the docstring.

### Risks
- **Blast radius of the frontend cut is well-contained** — no backend endpoint changes, so restoring audio/midi is genuinely a 2-line revert as advertised. Good demo-day hygiene.
- **Proxy path is the new hot path for every demo download** — if TuneChat goes down mid-demo, every download chip 502s. Consider a brief user-facing message ("Upstream render service unavailable") in the frontend error handler for 502s specifically, so it doesn't look like an Oh Sheet bug on stage.
- **No integration test with a real TuneChat job** — the manual QA checkbox (`Manual on qa`) is the only end-to-end verification. Fine given demo urgency, but worth running before the demo window opens.

### Verdict
**Approve for demo** with the SSRF allowlist (#1) as a recommended follow-up for post-demo hardening. Everything else is polish — the core change is small, well-scoped, and has good test coverage on the failure modes that matter (local fallback + 502).

jack0lantern (approved):

robin-raq:
## Review response — addressed vs deferred

Thanks for the thorough review. Validated each of the 8 points against the live code and split them into ship-now vs post-demo polish. Summary below.

### Addressed in commit `dd650d5`

**#1 — SSRF: URL trusted blindly**
Added `_is_allowed_proxy_target(url)` — strict `(scheme, hostname, port)` equality against `settings.tunechat_url` before any network I/O. Mismatch → 502 "Upstream artifact host is not allowed." Default-port normalization handles `http://localhost` vs `http://localhost:80` correctly.

Two new tests lock the guard in:
- `test_artifact_refuses_proxy_to_foreign_host` — simulates a TuneChat JSON response pointing at `169.254.169.254` (AWS IMDS). Uses an `ExplodeClient` that `raise AssertionError` on any `.get()`, so the test fails loudly if the guard is ever bypassed. Stronger than `assert_not_called` because it catches control-flow regressions, not just call counts.
- `test_artifact_proxy_scheme_mismatch_is_rejected` — https-target vs http-configured → blocked. Prevents downgrade/upgrade swaps that would bypass a host-only check.

**#8 — Log line could leak future signed-URL tokens**
Added a `# NOTE:` next to the warning log flagging the forward-risk: today TuneChat returns path-only URLs so logging the full URL is safe, but if TuneChat ever adopts signed URLs (`?token=...`, presigned-S3 style) the log becomes a credential leak. When that happens, strip the query before logging.

### Deferred as post-demo polish

All valid, all tracked, none blocking:

| # | Item | Why deferred |
|---|---|---|
| **2** | Module-level `httpx.Client` singleton | Pure perf. Per-request TLS handshake cost is invisible at demo scale (~10 downloads total). Worth doing before any traffic spike post-launch. |
| **3** | `client.stream()` + `StreamingResponse` | Current artifacts are <1 MB each (PDF + MusicXML + MIDI). Memory cliff only materializes if TuneChat ever serves large files. Drop-in upgrade, but no demo value. |
| **4** | Test for `?inline=true` on the proxy path | Code path is correct — line 79 sets `disposition` and line 101 applies it in the proxy branch. Just a test gap. ~5-line test. |
| **5** | `_install_tunechat_job` pokes `dependency_overrides` internals | Works today; fragile if `get_job_manager` becomes async. Right fix is a conftest fixture exposing the live `JobManager`. |
| **6** | Three `FakeClient` definitions could share a factory | Pure style. Factoring helps the next person touching these tests; doesn't change behavior. |

### Pushed back on

**#7 — `.body` sentinel in `test_artifact_idle_title_fallback`**
Reviewer worried `.body` is too generic and would break silently on a rename. Fair concern, but the test's *intent* is "still renders a form body even for unknown sources" — `.body` is literally the right assertion. All 5 phase bodies in `views.js` use the `.body` class. Using a phase-specific sentinel (e.g. the YouTube URL input) would couple the test to the YouTube-fallback detail, which is exactly what PR #83's picker simplification already made implicit. Leaving as-is; open to revisiting if the rename-risk concern outweighs the semantic fit.

### Risk notes from the review (kept as-is)

- **"Blast radius of the frontend cut is well-contained"** — agree, backend endpoints untouched, 2-line revert as advertised.
- **"If TuneChat goes down mid-demo, every download chip 502s"** — true. Deferred adding a user-facing "Upstream render service unavailable" message; existing `state.js` error handling will still surface *something*. If this bites on stage we'll patch post-demo.
- **"No integration test with a real TuneChat job"** — unit coverage is good, but the smoke test against a real qa deployment is the only end-to-end verification. Will do before the demo window opens.

CI should be green now. 12/12 backend artifact tests + lint clean.

PR #96, opened 2026-04-20
fix(engrave): fall through to local pipeline when TuneChat fails

jack0lantern:
## Summary
- **title_lookup jobs now fall through to the local Oh Sheet pipeline** (transcribe → arrange → humanize → engrave) when TuneChat is disabled or returns None, instead of crashing with a `RuntimeError`
- A lower-quality local result is better than a user-facing error
- Updated tests to assert successful fallthrough behavior

## Test plan
- [x] `test_title_lookup_falls_through_to_local_engrave_when_tunechat_disabled` — verifies fallthrough when TuneChat is off
- [x] `test_title_lookup_falls_through_to_local_engrave_when_tunechat_returns_none` — verifies fallthrough when TuneChat returns None

### DIFF
--- a/backend/jobs/runner.py
+++ b/backend/jobs/runner.py
@@ -429,16 +429,16 @@ class PipelineRunner:
                         score_dict = refined["payload"]
 
                 elif step == "engrave":
-                    # title_lookup jobs must resolve upstream (TuneChat) and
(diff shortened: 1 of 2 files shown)


lramosve (approved):
Looks good, ship it!!

PR #102, opened 2026-04-23
feat(about): add About page to frontend-v2 with team link

lramosve:
## Summary

Adds an About page to frontend-v2 (the live Vite SPA). PR #101 added the About screen to the old Flutter frontend which is no longer deployed.

### Changes

- **frontend-v2/assets/about.html** - Standalone page with team bios, LinkedIn links, and badges, styled to match the kawaii sticker aesthetic. Lives in Vite publicDir so it copies as-is to dist/about.html.
- **frontend-v2/index.html** - Added "About the team" link in the footer next to "Powered by AI".

### How it works

The backend serves everything from /app/static/ via FastAPI StaticFiles(html=True), so /about.html resolves automatically. No backend or Vite config changes needed.

## Test plan

- [ ] Deploy to QA and verify "About the team" link appears in the footer
- [ ] Click the link and verify the About page renders with all 5 team members
- [ ] Click "Back to Oh Sheet!" and verify it returns to the main page
- [ ] Verify LinkedIn links open correctly

### DIFF
--- a/frontend-v2/assets/about.html
+++ b/frontend-v2/assets/about.html
@@ -4,18 +4,20 @@
 <meta charset="utf-8" />
 <meta name="viewport" content="width=device-width, initial-scale=1" />
 <title>About - Oh Sheet!</title>
+<meta name="description" content="Meet the team behind Oh Sheet! — five engineers who built an AI pipeline that turns any song into playable piano sheet music." />
+<meta property="og:title" content="About - Oh Sheet!" />
+<meta property="og:description" content="Meet the team behind Oh Sheet! — five engineers who built an AI pipeline that turns any song into playable piano sheet music." />
 <meta name="theme-color" content="#FDF4E8" />
 <link rel="icon" type="image/svg+xml" href="/favicon.svg" />
 <link rel="preconnect" href="https://fonts.googleapis.com">
 <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin>
-<link href="https://fonts.googleapis.com/css2?family=Roboto:wght@400;500;600;700;800;900&family=Material+Symbols+Rounded:opsz,wght,FILL,GRAD@24,400,0,0" rel="stylesheet">
+<link href="https://fonts.googleapis.com/css2?family=Roboto:wght@400;500;600;700;800;900&family=Material+Symbols+Rounded:opsz,wght,FILL,GRAD@24,400,0,0&icon_names=arrow_back,groups_2,link,schedule,school" rel="stylesheet">
+<!-- Tokens are redefined here rather than shared with src/style.css
+     because the two pages use slightly different palette values (e.g.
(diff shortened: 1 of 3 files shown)


robin-raq (changes requested):
## Review Summary

Solid architectural choice — `publicDir: "assets"` in `vite.config.js` makes `about.html` copy through to `dist/` verbatim, and the existing `StaticFiles(html=True)` mount handles `/about.html` with no backend changes. Requesting changes primarily for one security item and a few a11y/maintainability nits.

### Must fix

- **Missing `rel="noopener noreferrer"` on all 5 `target="_blank"` LinkedIn links** (`about.html:94, 102, 110, 118, 126`). Browsers default to `noopener` for `target=_blank` nowadays, but explicit is the audited standard — and `noreferrer` still matters (prevents LinkedIn seeing the source URL as a referrer).

### Should fix

- **Design tokens duplicated inline** (`about.html:12–16`): `--bg`, `--ink`, `--primary`, etc. are redefined in `<style>` instead of shared with `src/style.css`. If the palette shifts, this page silently drifts. A publicDir file can't `@import` from `/src/` (Vite hashes that path in prod), so the fix is a standalone `assets/about.css` — or better, a shared `assets/tokens.css` both pages `<link>`.
- **`"Meet the team"` is a `<div class="section-title">`** (`about.html:90`), not a heading. Screen readers skip it in heading navigation — should be `<h2>`.
- **Decorative icons lack `aria-hidden="true"`**: the `.material-symbols-rounded` glyphs (arrow_back, link, groups_2, schedule, school) get announced as their ligature text by some screen readers. Cheap to fix now.
- **Inline styles should be classes**: per-card accent colors (`style="color:#FFB347"`, `style="background:#FFB347"`) and the footer link styles in `index.html:59`. With 4 palette colors across 5 members, a `.card--orange` / `.card--green` / etc. modifier cleanly scopes dot + link color together.

### Nice to have

- **Subset the icon font**: Material Symbols Rounded ships ~300KB+ for the full set. Add `&icon_names=arrow_back,link,groups_2,schedule,school` to the Google Fonts URL.
- **SEO**: no `<meta name="description">` or OG tags — the page won't get a meaningful search snippet. `index.html` has one.
- **Duplicate accent color**: Jack Jiang and Ross Kuehl both use `#FFB347`. With 4 colors for 5 members, one repeats — flagging in case it's unintended.

### What's already good

- Vite `publicDir` usage verified and correct.
- No XSS surface (static content, no user input).
- No backend changes needed — reuses existing `StaticFiles(html=True)`.
- Manual test plan is appropriate for a static page.
- Footer link contrast over `--bg` is fine.

### Risk

Low overall. The `rel="noopener noreferrer"` fix is the only hard blocker; the inline `<style>` block is the item most likely to cause future maintenance friction, so worth extracting while the PR is open.

